PR #94, opened 2023-03-21
feat(notification): add new bq-notification component

urosendv:


packages/bee-q/src/components/notification/bq-notification.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+    return (
+      <Host aria-hidden={this.isHidden} hidden={this.isHidden}>
+        <div
+          class="bg-white notification-shadow notification-radius inline-block w-auto px-5 py-4"
+          id="notification-main-holder"
+        >
+          {this.showIcon && (
+            <div class="mr-2 inline-block text-left align-top" id="notification-icon-holder">
+              <bq-icon name={this.getNotificationIconName()} color={this.subjectColor} size="18"></bq-icon>
```
dgonzalezr, line 130: According to the design specs, the icon should be 24px (default size)

CleanShot 2023-03-23 at 10 25 10

packages/bee-q/src/components/notification/bg-notification-types.ts
```diff
@@ -0,0 +1 @@
+export const NOTIFICATION_TYPES = ['info', 'success', 'warning', 'error'] as const;
```
dgonzalezr, line 1: I would recommend adding a `default` type as well, were the icon would not take any specific color.

```suggestion
export const NOTIFICATION_TYPE = ['default', 'info', 'success', 'warning', 'error'] as const;
export type TNotificationType = (typeof NOTIFICATION_TYPE)[number];
```

and then you would be able to set in the component implementation:

```ts
@Prop({ reflect: true }) type: TNotificationType = 'default';
```

packages/bee-q/src/components/notification/bq-notification.tsx
```diff
@@ -0,0 +1,169 @@
...
+  render() {
+    const styles = {
+      ...(this.subjectColor && { color: getColorCSSVariable(this.subjectColor) }),
+    };
+
+    return (
+      <Host aria-hidden={this.isHidden} hidden={this.isHidden}>
+        <div
+          class="bg-white notification-shadow notification-radius inline-block w-auto px-5 py-4"
+          id="notification-main-holder"
```
dgonzalezr, line 126: Please remove the usage of `id` attributes if you are not using them for targeting elements from JS

packages/bee-q/src/components/notification/bq-notification.tsx
```diff
@@ -0,0 +1,169 @@
...
+              <bq-icon name={this.getNotificationIconName()} color={this.subjectColor} size="18"></bq-icon>
+            </div>
+          )}
+          {!this.showIcon && (
+            <div class="avatar-slot-holder inline-block text-left align-top" id="notification-avatar-holder">
+              <slot name="avatar" />
+            </div>
+          )}
+          <div class="max-w-x inline-block text-left align-top">
+            <div class="title-font font-semibold" style={styles} id="subject">
```
dgonzalezr, line 139: Styles for overriding component's custom properties should be set in the `<Host>` instead.

packages/bee-q/src/components/notification/bq-notification.tsx
```diff
@@ -0,0 +1,169 @@
...
+              <slot name="description" />
+            </div>
+            <div class={{ 'mt-3': this.hasBtn }}>
+              <span class={{ 'mr-2': this.hasBtn }} ref={(spanElem) => (this.buttonElem = spanElem)}>
+                <slot name="first-button" onSlotchange={this.handleSlotChange} />
+              </span>
+              <span>
+                <slot name="second-button" />
+              </span>
+            </div>
```
dgonzalezr, line 152: We just need a footer slot here. Is up to the consumer to provide the number of actions needed.
The parent container of the slot should handle spacing between the elements inside, something that you can do with flex and gap 🙂

packages/bee-q/src/components/notification/bq-notification.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+  render() {
+    const styles = {
+      ...(this.subjectColor && { color: getColorCSSVariable(this.subjectColor) }),
+    };
+
+    return (
+      <Host aria-hidden={this.isHidden} hidden={this.isHidden}>
+        <div
+          class="bg-white notification-shadow notification-radius inline-block w-auto px-5 py-4"
```
dgonzalezr, line 125: Padding for the component content should 16px all over according to Design specs:
CleanShot 2023-03-23 at 12 14 12

I would recommend setting a component custom property for that:
```
--bq-notification--padding: theme('padding.4');
```

and then you can use it in JSX as:
```
class="p-[var(--bq-notification--padding)]"
```

packages/bee-q/src/components/notification/scss/bq-notification.variables.scss
dgonzalezr, line 1: We should also set a min-width custom property:
```css
--bq-notification--min-width: 320px;
```
according to Design specs is the min-width of the notification

packages/bee-q/src/components/notification/bq-notification.tsx
urosendv: Fixed this. I was using id's for E2E testing... I will also need to change my tests but that will be in another commit because I also need to add some other tests.

packages/bee-q/src/components/notification/bq-notification.tsx
```diff
@@ -0,0 +1,188 @@
...
+  @State() private isHidden = true;
+
+  // Public Property API
+  // ========================
+
+  /** Type of Notification */
+  @Prop({ reflect: true }) type: TNotificationType = 'default';
+
+  /** Set property if you want Notification icon to be shown. */
+  @Prop({ reflect: true }) showIcon: boolean;
```
dgonzalezr, line 41: We should use a `hideIcon` prop instead, normally, things that are intended to be shown by default should not be declared as true, and, on the contrary, to set a prop to hide them if needed.

```suggestion
  /** If true, the notification icon won't be shown */
  @Prop({ reflect: true }) hideIcon: boolean;
```

packages/bee-q/src/components/notification/bq-notification.tsx
```diff
@@ -0,0 +1,188 @@
...
+    if (type === 'success') {
+      icon = 'check-circle';
+    }
+
+    if (type === 'error') {
+      icon = 'x-circle';
+    }
+
+    return icon;
+  };
```
dgonzalezr, line 114: Is it possible to use a `switch` method instead? Something like:

```tsx
  private get iconName() {
    if (this.hideIcon) return;

    switch (this.type) {
      case 'error':
        return 'x-circle';
      case 'success':
        return 'check-circle';
      case 'warning':
        return 'warning-circle';
      default:
        return 'info';
    }
  }
```

packages/bee-q/src/components/notification/bq-notification.tsx
```diff
@@ -0,0 +1,188 @@
...
+          {this.showIcon && (
+            <div class="mr-2 inline-block text-left align-top" part="icon">
+              <bq-icon name={this.getNotificationIconName()} color={this.getNotificationColor()} size="24"></bq-icon>
+            </div>
+          )}
+          {!this.showIcon && (
+            <div class="avatar-slot-holder inline-block text-left align-top" part="avatar">
+              <slot name="avatar" />
+            </div>
+          )}
```
dgonzalezr, line 161: I don't think we need a separate slot for avatar and icon, usually, if the consumer would like to show an avatar instead of an icon, the icon slot should be the one to use since the intention is the same. That said, maybe something like this could work 👇🏼 

```jsx
{/* ICON */}
{!this.hideIcon && (
  <div
    class={{
      'notification--icon mr-2 inline-block text-left align-top': true,
      [`color-${this.type}`]: true, // The icon color will be based on the type (info, success, warning, error)
    }}
  >
    {!this.hasCustomIcon ? <bq-icon name={this.iconName} /> : <slot name="icon" />}
  </div>
)}
```

By adding a property `hasCustomIcon` we can provide the flexibility to use the icon slot to display something different from the predefined icons, even an avatar.

packages/bee-q/src/components/notification/bq-notification.tsx
```diff
@@ -0,0 +1,188 @@
...
+    const colors = {
+      default: 'ui--inverse-active',
+      success: 'ui--success',
+      info: 'ui--brand',
+      warning: 'ui--warning',
+      error: 'ui--danger',
+    };
+
+    return isDefined(this.subjectColor) ? this.subjectColor : colors[type];
+  };
```
dgonzalezr, line 127: We can leverage on CSS to set the icon color based on the notification type:

If we can have something like:

```jsx
{/* ICON */}
{!this.hideIcon && (
  <div
    class={{
      'notification--icon mr-2 inline-block text-left align-top': true,
      [`color-${this.type}`]: true, // The icon color will be based on the type (info, success, warning, error)
    }}
  >
    {!this.hasCustomIcon ? <bq-icon name={this.iconName} /> : <slot name="icon" />}
  </div>
)}
```
then we can do the following with SCSS
```scss
/**
 * Set the nocitication icon color based on the type value selected
 */
.notification--icon.color {
  &-default {
    @apply text-text-primary;
  }

  &-info {
    @apply text-text-brand;
  }

  &-error {
    @apply text-text-danger;
  }

  &-success {
    @apply text-text-success;
  }

  &-warning {
    @apply text-text-warning;
  }
}
```

packages/bee-q/src/components/notification/bq-notification.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+  // render() function
+  // Always the last one in the class.
+  // ===================================
+
+  render() {
+    const styles = {
+      ...(this.subjectColor && { color: getColorCSSVariable(this.subjectColor) }),
+    };
```
dgonzalezr, line 144: There's no need to set the subject color in the component since the title is provided by the consumer, and the text color can be set on integration.

packages/bee-q/src/components/notification/bq-notification.tsx
```diff
@@ -0,0 +1,188 @@
...
+  // Reference to host HTML element
+  // ===================================
+
+  @Element() el!: HTMLBqNotificationElement;
+
+  // State() variables
+  // Inlined decorator, alphabetical order
+  // =======================================
+
+  @State() private isHidden = true;
```
dgonzalezr, line 32: We could use a mutable `isOpen` property instead of state.

```tsx
/** If true, the notification will be shown */
@Prop({ reflect: true, mutable: true }) isOpen = false;
```

packages/bee-q/src/components/notification/bq-notification.tsx
```diff
@@ -0,0 +1,188 @@
...
+  // Always the last one in the class.
+  // ===================================
+
+  render() {
+    const styles = {
+      ...(this.subjectColor && { color: getColorCSSVariable(this.subjectColor) }),
+    };
+
+    return (
+      <Host style={styles} aria-hidden={this.isHidden} hidden={this.isHidden}>
```
dgonzalezr, line 147: For HTML attributes that receive "true" or "false" values we should do something like:

```jsx
<Host
  class={{ '!hidden': !this.isOpen }}
  aria-hidden={!this.isOpen ? 'true' : 'false'}
  hidden={!this.isOpen ? 'true' : 'false'}
>
```

packages/bee-q/src/components/notification/bq-notification.tsx
```diff
@@ -0,0 +1,188 @@
...
+            <div class="w-notification-close inline-block text-right align-top" part="close-icon">
+              <bq-icon
+                name="x"
+                color="icon--primary"
+                size="14"
+                class="cursor-pointer"
+                onClick={this.hideNotification}
+              ></bq-icon>
+            </div>
+          )}
```
dgonzalezr, line 183: I believe that the close button icon should be a bit up in the HTML layout hierarchy:

```jsx
<Host
  class={{ '!hidden': !this.isOpen }}
  aria-hidden={!this.isOpen ? 'true' : 'false'}
  hidden={!this.isOpen ? 'true' : 'false'}
>
  <div
    class="relative inline-flex min-w-[var(--bq-notification--min-width)] items-start rounded-[var(--bq-notification--border-radius)] bg-bg-primary p-[var(--bq-notification--padding)] shadow-m"
    part="base"
  >
    {/* CLOSE BUTTON */}
    {!this.disableClose && (
      <bq-button
        class="notification--close absolute right-5"
        appearance="text"
        size="small"
        onClick={() => this.hide()}
        part="btn-close"
      >
        <bq-icon name="x" />
      </bq-button>
    )}
    {/* ICON */}
    {!this.hideIcon && (
      <div
        class={{
          'notification--icon mr-2 inline-block text-left align-top': true,
          [`color-${this.type}`]: true, // The icon color will be based on the type (info, success, warning, error)
        }}
      >
        {!this.hasCustomIcon ? <bq-icon name={this.iconName} /> : <slot name="icon" />}
      </div>
    )}
    {/* CONTENT */}
    <div class="flex flex-col items-start gap-4" part="content">
      {/* TITLE */}
      <div class="title-font font-semibold leading-large" part="title">
        <slot />
      </div>
      {/* BODY */}
      <div class="text-s leading-regular" part="body">
        <slot name="body" />
      </div>
      {/* FOOTER */}
      <div class="flex items-start gap-2" part="footer">
        <slot name="footer" />
      </div>
    </div>
  </div>
</Host>
```

That will guarantee that moving focus around with the Tab key will get the right order.

packages/bee-q/src/components/notification/bq-notification.tsx
```diff
@@ -0,0 +1,176 @@
...
+  @Element() el!: HTMLBqNotificationElement;
+
+  // State() variables
+  // Inlined decorator, alphabetical order
+  // =======================================
+
+  // Public Property API
+  // ========================
+
+  /** If rue, the close button at the top right of the notification won't be shown */
```
endv-bogdanb, line 38:
```suggestion
  /** If true, the close button at the top right of the notification won't be shown */
```

endv-bogdanb (approved):

PR #97, opened 2023-03-22
feat(dialog): add dialog component

raresherta:


packages/bee-q/src/components/dialog/bq-dialog.tsx
```diff
@@ -0,0 +1,145 @@
...
+          }}
+        >
+          <header class="bq-header">
+            <div class="bq-placeholder-info">
+              <div class="bq-info">
+                <slot name="info" />
+              </div>
+            </div>
+            <div class="flex flex-col pl-4">
+              <h3>
```
endv-bogdanb, line 115: Don't think we need the h3 here. The user will provide this as part of slot.
```suggestion
              
```

endv-bogdanb:
Check #269.

PR #112, opened 2023-03-28
Feature/toast component new

marinakostic:
feat(toast): add new toast component


packages/bee-q/src/components/toast/bq-toast.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  // Local methods
+  // Internal business logic.
+  // These methods cannot be called from the host element.
+  // =======================================================
+
+  // render() function
+  // Always the last one in the class.
+  // ===================================
+  private getColorAndIcon = () => {
```
endv-bogdanb, line 86: Please move this one between 
```text
 // Local methods
  // Internal business logic.
  // These methods cannot be called from the host element.
  // =======================================================

  // render() function
  // Always the last one in the class.
```

packages/bee-q/src/components/toast/bq-toast.tsx
```diff
@@ -0,0 +1,135 @@
...
+    const type = this.type;
+    const textColor = this.textColor;
+    const defaultColors = {
```
endv-bogdanb, line 89: It can be split into two functions
  private getColorAndIcon = () => {
```
private get iconColor(){}

private get icon(){}
```

this way you can just call it directly as property even though is computed `this.iconColor `

packages/bee-q/src/components/toast/bq-toast.tsx
```diff
@@ -0,0 +1,135 @@
...
+  // Public methods API
+  // These methods are exposed on the host element.
+  // Always use two lines.
+  // Public Methods must be async.
+  // Requires JSDocs for public API documentation.
+  // ===============================================
+
+  /** Trigers function to show toast */
+  @Method()
+  async showToast() {
```
endv-bogdanb, line 70: We should provide a method to hide toast if the user doesn't want it to `autoclose`.

packages/bee-q/src/components/toast/bq-toast.tsx
```diff
@@ -0,0 +1,135 @@
...
+  /** Text color of Toast */
+  @Prop({ reflect: true }) textColor: string;
+  /** Should show icon of Toast */
+  @Prop({ reflect: true }) showIcon = false;
+  /** Should hide Toast after period of time */
+  @Prop({ reflect: true }) autoCloseTime: number;
+
+  // Prop lifecycle events
+  // =======================
+  @Watch('type')
```
endv-bogdanb, line 42: make sure the timer is always `>=0`
```typescript

      this.autoCloseTime = Math.max(0, this. autoCloseTime);

```

```suggestion
  @Watch('type')
  @Watch('autoCloseTime')
```

packages/bee-q/src/components/toast/bq-toast.tsx
```diff
@@ -0,0 +1,135 @@
...
+  // Public Property API
+  // ========================
+  /** Type of Toast */
+  @Prop({ reflect: true, mutable: true }) type: TToastType = 'default';
+  /** Text color of Toast */
+  @Prop({ reflect: true }) textColor: string;
+  /** Should show icon of Toast */
+  @Prop({ reflect: true }) showIcon = false;
+  /** Should hide Toast after period of time */
+  @Prop({ reflect: true }) autoCloseTime: number;
```
endv-bogdanb, line 38: I would make this property accept a null as well and default it to `5000ms` if null is provided it will stay open until hide method is called

packages/bee-q/src/components/toast/bq-toast.tsx
```diff
@@ -0,0 +1,135 @@
...
+  // Always use two lines.
+  // Public Methods must be async.
+  // Requires JSDocs for public API documentation.
+  // ===============================================
+
+  /** Trigers function to show toast */
+  @Method()
+  async showToast() {
+    this.shouldShowToast = true;
+    const timeout = !this.autoCloseTime ? 5000 : this.autoCloseTime;
```
endv-bogdanb, line 72: We need the timeout as property. Maybe check the usage of `debounce` util, I think we can leverage that to auto close the toaster. Also it provides a `cancel` that can be called if the user decides to close it faster.

endv-bogdanb:
I would add `bqClose` event, that's triggered when the toaster is closed.

endv-bogdanb:
Close in favour of #301

PR #246, opened 2023-05-08
feature(Dropdown): add new `bq-panel` and `bq-dropdown` components

HamudeHomsi:


packages/bee-q/src/components/dropdown/__tests__/bq-dropdown.e2e.ts
dgonzalezr, line 1: Please try to add more tests for functionalities, e.g:
* when it should be visible or not
* if it has rounded corners
* etc...

packages/bee-q/src/components/dropdown/_storybook/bq-dropdown.stories.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+  const avatarTrigger = html` <bq-avatar
+    alt-text="User profile"
+    image="https://images.unsplash.com/photo-1524593689594-aae2f26b75ab?ixlib=rb-1.2.1&amp;ixid=eyJhcHBfaWQiOjEyMDd9&amp;auto=format&amp;fit=crop&amp;w=1000&amp;q=80"
+    label="Avatar component label"
+    initials=""
+    shape="circle"
+    size="medium"
+    slot="trigger"
+  ></bq-avatar>`;
```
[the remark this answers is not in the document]
HamudeHomsi: Yes, will refactor this today.

packages/bee-q/src/components/dropdown/_storybook/bq-dropdown.stories.tsx
```diff
@@ -0,0 +1,169 @@
...
+  return html`
+    <bq-dropdown
+      panel-distance=${args['panel-distance']}
+      panel-placement=${args['panel-placement']}
+      panel-open=${args['panel-open']}
+      panel-scrollbar=${args['panel-scrollbar']}
+      @bqPanelChange=${args.bqPanelChange}
+    >
+      <!-- TRIGGER ELEMENT -->
+      ${args.children}
```
dgonzalezr, line 52: Perhaps, calling this arg `trigger` will be more semantic, what do you think? 🤔

packages/bee-q/src/components/dropdown/_storybook/bq-dropdown.stories.tsx
```diff
@@ -0,0 +1,169 @@
...
+        <bq-button appearance="secondary" size="medium" type="button" variant="standard" slot="trigger">
+          <span class="flex items-center gap-1">
+            <span>Dropdown</span>
+            <bq-icon name="caret-down"></bq-icon>
+          </span>
+        </bq-button>
+      `,
+    })}
+  `,
+};
```
dgonzalezr, line 98:
```suggestion
export const Default: Story = {
  render: (args: Args) => html`
    ${Template({
      ...args,
      trigger: html`
        <bq-button appearance="secondary" size="medium" type="button" variant="standard" slot="trigger">
          <span class="flex items-center gap-1">
            <span>Dropdown</span>
            <bq-icon name="caret-down"></bq-icon>
          </span>
        </bq-button>
      `,
    })}
  `,
};
```

packages/bee-q/src/components/dropdown/_storybook/bq-dropdown.stories.tsx
```diff
@@ -0,0 +1,169 @@
...
+export default meta;
+
+type Story = StoryObj;
+
+const Template = (args: Args) => {
+  return html`
+    <bq-dropdown
+      panel-distance=${args['panel-distance']}
+      panel-placement=${args['panel-placement']}
```
dgonzalezr, line 46: You can use a direct return here:

```suggestion
const Template = (args: Args) => html`
    <bq-dropdown
      panel-distance=${args['panel-distance']}
      panel-placement=${args['panel-placement']}
```

packages/bee-q/src/components/dropdown/bq-dropdown.tsx
```diff
@@ -0,0 +1,164 @@
...
+  /**
+   * If true, panel is visible.
+   * You can toggle this attribute to show/hide the panel.
+   */
+  @Prop({ reflect: true }) panelOpen?: boolean = false;
+
+  /**
+   * Determines whether the scrollbar is visible or hidden within the panel.
+   */
+  @Prop({ reflect: true }) panelScrollbar?: boolean = false;
```
dgonzalezr, line 49: I don't think we need a property to set scroll in the panel. Instead, I would recommend having a `panelHeight` property, and if the amount of options is more than the available height, the panel body should scroll automatically.

packages/bee-q/src/components/dropdown/bq-dropdown.tsx
```diff
@@ -0,0 +1,164 @@
...
+  @Prop({ reflect: true }) panelDistance?: number = 4;
+
+  /** Position of the panel */
+  @Prop({ reflect: true }) panelPlacement?: FloatingUIPlacement = 'bottom-start';
+
+  /**
+   * If true, panel is visible.
+   * You can toggle this attribute to show/hide the panel.
+   */
+  @Prop({ reflect: true }) panelOpen?: boolean = false;
```
dgonzalezr, line 44: Can we use just `open`, `distance`, and `placement` instead?

packages/bee-q/src/components/dropdown/bq-dropdown.tsx
```diff
@@ -0,0 +1,164 @@
...
+  /** Distance (px) between the panel and the trigger element. */
+  @Prop({ reflect: true }) panelDistance?: number = 4;
+
+  /** Position of the panel */
+  @Prop({ reflect: true }) panelPlacement?: FloatingUIPlacement = 'bottom-start';
+
+  /**
+   * If true, panel is visible.
+   * You can toggle this attribute to show/hide the panel.
+   */
```
dgonzalezr, line 43:
```suggestion
  /** If true, panel is visible. You can toggle this attribute to show/hide the panel. */
```

packages/bee-q/src/components/panel/bq-panel.tsx
```diff
@@ -0,0 +1,147 @@
...
+  @Watch('placement')
+  onPropChange() {
+    this.floatingUI.init({
+      placement: this.placement,
+      distance: this.distance,
+      sameWidth: false,
+      strategy: 'fixed',
+      skidding: 0,
+    });
+  }
```
dgonzalezr, line 68: I'm not truly convinced that this logic should belong to the panel. My understanding is that the panel should only handle the styles (shadow, stroke, spacing) and the click/select of the options it has. It should be up to the dropdown itself how to handle showing/hiding the panel, distance, and so on 😉

HamudeHomsi: I think that the panel should handle the init of the panel, in our case `initFloatingUI()`, `toggle` panel, and `destroy` panel. Even though these methods are part of the `bq-panel`, they are called from the parent comp, so the control is at the parent level, `bq-dropdown`.
The `bq-panel` component should only handle the styles, as you mentioned, and the create, update floatingUI, "dumb" component. The handle of the selected option is at the `bq-dropdown` level.
Also, in the case of `bq-select`, I think this behavior should follow the same principle. Otherwise we would have to repeat the same code for init, update, destroy.
`bq-select` would be the parent component and have access to `bq-panel` public methods: `setTriggerElement()`, `togglePanel()`.
What do you think?

packages/bee-q/src/components/dropdown/bq-dropdown.tsx
HamudeHomsi: Agree. Will do!

packages/bee-q/src/components/panel/bq-panel.tsx
dgonzalezr: You are ending up duplicating properties to pass from the dropdown parent to the panel component, which I don't think to be ideal, but it could be that maybe I'm missing something. I'll play a bit with your solution in y local env 😉

dgonzalezr: After playing a bit, I agree that we can keep some logic in the `bq-panel`, but there is room for some improvements. I'll provide the relevant changes proposal in the block code related 😉

packages/bee-q/src/components/panel/bq-panel.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+  /** If true, panel is visible.
+   * You can toggle this attribute to show/hide the panel.
+   */
+  @Prop({ reflect: true }) open?: boolean = false;
+
+  /** If true, the scrollbar is visible.
+   * You can toggle this attribute to show/hide the scrollbar.
+   */
+  @Prop({ reflect: true }) scrollbar?: boolean = false;
```
dgonzalezr, line 48: We need to add all available options by FloatingUI, also **we can pass the trigger element through a property instead** of using a method for it.

```suggestion
  /** Represents the distance (gutter or margin) between the panel and the trigger element. */
  @Prop({ reflect: true }) distance?: number = 4;

  /** If true, the panel will be visible. */
  @Prop({ reflect: true, mutable: true }) open?: boolean = false;

  /** Position of the panel */
  @Prop({ reflect: true }) placement?: FloatingUIPlacement = 'bottom';

  /** The trigger element for the panel */
  @Prop() triggerElement?: HTMLElement;

  /** Whether the panel should have the same width as the trigger element */
  @Prop({ reflect: true }) sameWidth?: boolean = false;

  /**  Represents the skidding between the panel and the trigger element. */
  @Prop({ reflect: true }) skidding?: number = 0;

  /** Defines the strategy to position the panel */
  @Prop({ reflect: true }) strategy?: 'fixed' | 'absolute' = 'fixed';
```

packages/bee-q/src/components/panel/bq-panel.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+  private initFloatingUI = () => {
+    this.floatingUI = new FloatingUI(this.trigger, this.panel, {
+      placement: this.placement,
+      distance: this.distance,
+      sameWidth: false,
+      strategy: 'fixed',
+      skidding: 0,
+    });
+  };
```
dgonzalezr, line 125: I will propose adding here only the necessary private methods to handle:
- options mapping
- open the panel
- closing the panel

Something like:

```suggestion
  private get options() {
    return {
      distance: this.distance,
      placement: this.placement,
      sameWidth: this.sameWidth,
      skidding: this.skidding,
      strategy: this.strategy,
    };
  }

  private showPanel() {
    this.floatingUI?.update();
  }

  private async hidePanel() {
    this.open = false;
  }
```

packages/bee-q/src/components/panel/bq-panel.tsx
```diff
@@ -0,0 +1,147 @@
...
+  onPropChange() {
+    this.floatingUI.init({
+      placement: this.placement,
+      distance: this.distance,
+      sameWidth: false,
+      strategy: 'fixed',
+      skidding: 0,
+    });
+  }
+
```
dgonzalezr, line 69: When passing the trigger element through a property, we can have a `@Watch` for it:

```js
  @Watch('triggerElement')
  onTriggerElementChange() {
    if (!this.triggerElement) return;

    this.floatingUI = new FloatingUI(this.triggerElement, this.panel, { ...this.options });
  }
```

PS: `this.options` is a private get that maps the FloatingUI options

packages/bee-q/src/components/panel/bq-panel.tsx
```diff
@@ -0,0 +1,147 @@
...
+          'bq-panel': true,
+          'hide-scrollbar': !this.scrollbar,
+        }}
+        ref={(el) => (this.panel = el)}
+        aria-hidden={!this.open}
+        hidden={!this.open}
+        part="base"
+      >
+        <slot />
+      </div>
```
dgonzalezr, line 144: No need for scrollbar handling as discussed before. Please notice that the `aria-*` attribute received a string, either `true` or `false` for boolean cases.

```suggestion
      <div
        class="bq-panel"
        ref={(el) => (this.panel = el)}
        aria-hidden={!this.open ? 'true' : 'false'}
        hidden={!this.open}
        part="panel"
      >
        <slot />
      </div>
```

packages/bee-q/src/components/panel/bq-panel.tsx
```diff
@@ -0,0 +1,147 @@
...
+   */
+  @Event() bqPanelVisibility: EventEmitter<boolean>;
+
+  // Component lifecycle events
+  // Ordered by their natural call order
+  // =====================================
+
+  disconnectedCallback() {
+    this.floatingUI?.destroy();
+  }
```
dgonzalezr, line 85: Handle trigger element on componentDidLoad(), if is not yet available the `@Watch` will exit doing nothing.

```suggestion
  componentDidLoad() {
    this.onTriggerElementChange();
  }

  disconnectedCallback() {
    this.floatingUI?.destroy();
  }
```

packages/bee-q/src/components/panel/bq-panel.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+  // Events section
+  // Requires JSDocs for public API documentation
+  // ==============================================
+
+  /** Handler to be called to check if the panel is open or closed.
+   * Will emit every time the state of the panel is changed.
+   */
+  @Event() bqPanelVisibility: EventEmitter<boolean>;
```
dgonzalezr, line 77: I don't know if we need this custom event (at least not momentarily).

packages/bee-q/src/components/panel/bq-panel.type.tsx
dgonzalezr, line 1: Since we're using `FloatingUIPlacement`, do we still need this file?

packages/bee-q/src/components/dropdown/bq-dropdown.tsx
dgonzalezr, line 49: We need to allow the same properties that `bq-panel` should receive from the parent:

```suggestion
  /** Represents the distance (gutter or margin) between the dropdown panel and the trigger element. */
  @Prop({ reflect: true }) distance?: number = 4;

  /** If true, the dropdown panel will be visible. */
  @Prop({ reflect: true, mutable: true }) open?: boolean = false;

  /** Position of the dropdown panel */
  @Prop({ reflect: true }) placement?: FloatingUIPlacement = 'bottom';

  /** Whether the dropdown panel should have the same width as the trigger element */
  @Prop({ reflect: true }) sameWidth?: boolean = false;

  /**  Represents the skidding between the dropdown panel and the trigger element. */
  @Prop({ reflect: true }) skidding?: number = 0;

  /** Defines the strategy to position the dropdown panel */
  @Prop({ reflect: true }) strategy?: 'fixed' | 'absolute' = 'fixed';
```

packages/bee-q/src/components/dropdown/bq-dropdown.tsx
```diff
@@ -0,0 +1,164 @@
...
+  // ==============
+
+  /** On click outside the panel */
+  @Listen('click', { target: 'document', passive: true })
+  onClickOutsidePanel(event: MouseEvent) {
+    // close the panel on click, except click within the panel or within the trigger element
+    if (!event.composedPath().includes(this.panelElement) && !event.composedPath().includes(this.trigger)) {
+      this.panelElement.open = false;
+    }
+  }
```
dgonzalezr, line 82: Since `bq-panel` receives an `open` property that internally handles when to show/hide the panel content, we can just do something like this:

```suggestion
  /** Listens for the 'click' event on the document object
   * and closes the dropdown panel if the click is outside the component. */
  @Listen('click', { target: 'document', passive: true })
  onClickOutside(event: MouseEvent) {
    if (!this.open) return;
    // Close when clicking outside of the close element
    const path = event.composedPath();
    if (!path.includes(this.el)) {
      this.open = false;
    }
  }
```

packages/bee-q/src/components/dropdown/bq-dropdown.tsx
```diff
@@ -0,0 +1,164 @@
...
+   * If the user press on Escape button
+   * or `<bq-dropdown>` loses focus
+   * then hide the panel
+   */
+  @Listen('keyup', { target: 'window', passive: true })
+  onEscape(event: KeyboardEvent) {
+    if (event.key === 'Escape' || (event.key === 'Tab' && !event.composedPath().includes(this.el))) {
+      this.panelElement.open = false;
+    }
+  }
```
dgonzalezr, line 94: We should skip if the panel is already closed.

```suggestion
  /**
   * Listens for the 'keyup' event on the window object
   * and closes the dropdown panel if the 'Escape' key or 'Tab' key outside the component is pressed.
   */
  @Listen('keyup', { target: 'window', passive: true })
  onEscape(event: KeyboardEvent) {
    if (!this.open) return;
    if (event.key === 'Escape' || (event.key === 'Tab' && !event.composedPath().includes(this.el))) {
      this.open = false;
    }
  }
```

packages/bee-q/src/components/dropdown/bq-dropdown.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+  @Listen('bqPanelVisibility', { passive: true })
+  onPanelVisibilityChange(event: CustomEvent) {
+    const isOpened: boolean = event.detail as boolean;
+
+    this.bqPanelChange.emit({ opened: isOpened });
+  }
```
dgonzalezr, line 101: Not sure that we need it at this moment 😕

packages/bee-q/src/components/dropdown/bq-dropdown.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+  @Listen('bqSelect', { passive: true })
+  onItemSelect() {
+    this.panelElement.open = false;
+  }
```
dgonzalezr, line 106:
```suggestion
  @Listen('bqSelect', { passive: true })
  onItemSelect() {
    this.open = false;
  }
```

packages/bee-q/src/components/dropdown/bq-dropdown.tsx
```diff
@@ -0,0 +1,164 @@
...
+  // =======================================================
+
+  private openPanel = (): void => {
+    this.panelElement?.togglePanel();
+    this.panelElement.open = !this.panelElement.open;
+  };
+
+  private initFloatingUIFromPanel = (): void => {
+    this.panelElement?.setTriggerElement(this.el);
+  };
```
dgonzalezr, line 127: Since the trigger element it will be passed through a property, we only need to handle the trigger element click to toggle the `open` property:

```suggestion
  private togglePanel = (): void => {
    this.open = !this.open;
  };
```

packages/bee-q/src/components/dropdown/bq-dropdown.tsx
```diff
@@ -0,0 +1,164 @@
...
+          open={this.panelOpen}
+          scrollbar={this.panelScrollbar}
+          ref={(el) => (this.panelElement = el)}
+          aria-labelledby="dropdown"
+          role="region"
+          part="panel"
+        >
+          <slot />
+        </bq-panel>
+      </div>
```
dgonzalezr, line 161: Some Tailwind CSS utilities can be added directly in the JSX, no need to use `.scss` for a couple of rules.

```suggestion
      <div class="bq-dropdown" part="base">
        {/* TRIGGER ELEMENT */}
        <div
          class="bq-dropdown__trigger block"
          ref={(el) => (this.trigger = el)}
          onClick={this.togglePanel}
          aria-expanded={this.open ? 'true' : 'false'}
          aria-haspopup="true"
          part="trigger"
        >
          <slot name="trigger" />
        </div>
        {/* PANEL */}
        <bq-panel
          class="bq-dropdown__panel"
          distance={this.distance}
          placement={this.placement}
          open={this.open}
          sameWidth={this.sameWidth}
          skidding={this.skidding}
          strategy={this.strategy}
          triggerElement={this.trigger}
          aria-labelledby="dropdown"
          role="region"
          part="dropdown"
          exportparts="panel"
        >
          <slot />
        </bq-panel>
      </div>
```

packages/bee-q/src/components/panel/scss/bq-panel.scss
```diff
@@ -0,0 +1,22 @@
...
+/*                                Panel styles                */
+/* -------------------------------------------------------------------------- */
+
+@import './bq-panel-variables';
+
+:host {
+  @apply block;
+}
+
+.bq-panel {
```
dgonzalezr, line 11: We can add animation-fade-in for the panel to animate when opened.

```
.bq-panel {
  @include animation-fade-in;

  ...
}
```

packages/bee-q/src/components/dropdown/bq-dropdown.tsx
HamudeHomsi: Actually this prop does not set scroll in the panel. The panel has a fixed height. If the amount of options is bigger than the height, by default, the panel is scrollable, but the scrollbar is hidden just like for `bq-side-menu`.
`panelScrollbar` will set the scrollbar to visible.
You can check the `With Scrollbar` story from Storybook.

dgonzalezr: Then it is clear we don't need this property 😉

packages/bee-q/src/components/panel/bq-panel.type.tsx
HamudeHomsi: This file exist because it's used in `bq-dropdown.stories.tsx`. I tried to use `FloatingUIPlacement` instead, but an error occurs because it is viewed as `value` and not as `type`.

endv-bogdanb: Try `import { type FloatingUIPlacement } from "..."`.

HamudeHomsi: It doesn't seem to work
image

HamudeHomsi: @endv-bogdanb suggested a solution. To keep the type file and enforce type on constant.
image

packages/bee-q/src/components/panel/__tests__/bq-panel.e2e.ts
```diff
@@ -0,0 +1,69 @@
...
+    `);
+
+    const button = await page.find('bq-button');
+    await button.click();
+
+    const dropdown = await page.find('bq-dropdown');
+    const panel = dropdown.shadowRoot.querySelector('bq-panel');
+
+    expect(panel).not.toHaveAttribute('aria-hidden');
+  });
```
dgonzalezr, line 41: I think this e2e test belongs to the dropdown rather than the panel, right? 🤔

packages/bee-q/src/components/dropdown/__tests__/bq-dropdown.e2e.ts
dgonzalezr, line 1: Can we add more tests for the dropdown?

packages/bee-q/src/components/panel/scss/bq-panel-variables.scss
```diff
@@ -0,0 +1,37 @@
...
+  --bq-panel--border-radius: theme('borderRadius.s');
+  --bq-panel--border-style: none;
+  --bq-panel--border-width: unset;
+
+  --bq-panel--box-shadow: theme('boxShadow.m');
+  --bq-panel--paddingY: theme('spacing.xs');
+  --bq-panel--padding-start: theme('spacing.xs');
+  --bq-panel--padding-end: theme('spacing.xs');
+
+  --bq-panel--height: 200px;
```
dgonzalezr, line 34: I would rather recommend setting the panel height relative to the view height, e.g:

```suggestion
  --bq-panel--height: 50vh;
```

packages/bee-q/src/components/dropdown/bq-dropdown.tsx
```diff
@@ -0,0 +1,160 @@
...
+          part="trigger"
+        >
+          <slot name="trigger" />
+        </div>
+        {/* PANEL */}
+        <bq-panel
+          class="bq-dropdown__panel"
+          distance={this.distance}
+          placement={this.placement}
+          open={this.open}
```
dgonzalezr, line 145: We could add a panelHeight property that allows user to override the dropdown panel height. Something like:

```js
  /** When set, it will override the height of the dropdown panel */
  @Prop({ reflect: true }) panelHeight?: string;

  render() {
    const style = {
      ...(this.panelHeight && { '--bq-panel--height': this.panelHeight }),
    };

    return (
      <div class="bq-dropdown" part="base">
        ...
        {/* PANEL */}
        <bq-panel
          style={style}
          class="bq-dropdown__panel"
```

packages/bee-q/src/components/dropdown/bq-dropdown.tsx
```diff
@@ -0,0 +1,160 @@
...
+  // ===============================================
+
+  // Local methods
+  // Internal business logic.
+  // These methods cannot be called from the host element.
+  // =======================================================
+
+  private togglePanel = (): void => {
+    this.open = !this.open;
+  };
```
dgonzalezr, line 120: Another good improvement will be adding a property that allows users to set whether the panel dropdown should stay open or not after selection. E.g:

```js
  /** If true, the panel will remain open after a selection is made. */
  @Prop({ reflect: true }) keepOpenOnSelect?: boolean = false;

  ...

  @Listen('bqSelect', { passive: true })
  onItemSelect() {
    if (this.keepOpenOnSelect) return;

    this.open = false;
  }
```

packages/bee-q/src/components/panel/scss/bq-panel.scss
```diff
@@ -0,0 +1,20 @@
...
+  @include animation-fade-in;
+  @include hide-scrollbar;
```
dgonzalezr, line 13: I would recommend not hiding the scrollbar, the users may not notice that there are other options down in the list... Please keep a line break between `@include` and `@apply`.

packages/bee-q/src/components/panel/bq-panel.type.tsx
```diff
@@ -0,0 +1,16 @@
...
+  'right',
+  'right-start',
+  'right-end',
+  'bottom',
+  'bottom-start',
+  'bottom-end',
+  'left',
+  'left-start',
+  'left-end',
+] as const;
```
dgonzalezr, line 16: Back to the need for PANEL_PLACEMENT, I see it is only used in the Storybook doc example. In that case, I would rather prefer to hardcode the value in the *.stories.tsx file than adding a useless type:

```js
  argTypes: {
    distance: { control: 'number' },
    placement: {
      control: { type: 'select' },
      options: [
        'top',
        'top-start',
        'top-end',
        'bottom',
        'bottom-start',
        'bottom-end',
        'right',
        'right-start',
        'right-end',
        'left',
        'left-start',
        'left-end',
      ],
    }
```

packages/bee-q/src/components/dropdown/__tests__/bq-dropdown.e2e.ts
```diff
@@ -0,0 +1,21 @@
+import { newE2EPage } from '@stencil/core/testing';
+
+describe('bq-dropdown', () => {
+  it('should render', async () => {
+    const page = await newE2EPage();
+    await page.setContent('<bq-dropdown></bq-dropdown>');
```
dgonzalezr, line 6: Please update all the e2e tests to use the shorthand for setting the page content:

```suggestion
    const page = await newE2EPage({
      html: `<bq-dropdown></bq-dropdown>`
    });
```

packages/bee-q/src/components/panel/scss/bq-panel-variables.scss
HamudeHomsi: Ok, but isn't `50vh` too much for a dropdown?

packages/bee-q/src/components/dropdown/bq-dropdown.tsx
HamudeHomsi: Like this!

dgonzalezr (approved):

PR #253, opened 2023-05-12
feature(option): add new `<bq-option>` component.

HamudeHomsi:


dgonzalezr:
# ☁️ Nx Cloud Report

CI is running/has finished running commands for commit 10853cfe9a78130fb7210c885d8acd5fdeeabb7f. As they complete they will appear below. Click to see the status, the terminal output, and the build insights.

[📂 See all runs for this branch](https://cloud.nx.app/branch?workspaceId=64007df8da9fd31ee8a4d06e&branchName=253,feature/option-component)

---


<details>
<summary>✅ Successfully ran 5 targets </summary>

- [`nx affected --target=e2e --base=9e8899437c29bb26269264e7347c10e4f5eafadc --head=10853cfe9a78130fb7210c885d8acd5fdeeabb7f --exclude=workspace --output-style=stream-without-prefixes -- --ci --code-coverage --runInBand`](https://cloud.nx.app/runs/YhoxXaXH7L)
- [`nx run bee-q:storybook-build --output-style=stream-without-prefixes`](https://cloud.nx.app/runs/bY1k0Eox9l)
- [`nx run bee-q:storybook-build-css --output-style=stream-without-prefixes`](https://cloud.nx.app/runs/7Zj785nBIX)
- [`nx run-many --target=build --exclude=workspace --output-style=stream-without-prefixes`](https://cloud.nx.app/runs/Obj3r2aklT)
- [`nx affected --target=test --base=9e8899437c29bb26269264e7347c10e4f5eafadc --head=10853cfe9a78130fb7210c885d8acd5fdeeabb7f --exclude=workspace --output-style=stream-without-prefixes -- --ci --code-coverage --runInBand`](https://cloud.nx.app/runs/VHSrmPLmFj)
</details>

---

Sent with 💌 from [NxCloud.](https://nx.app)

packages/bee-q/src/components/option/bq-option.tsx
```diff
@@ -0,0 +1,178 @@
...
+  @Element() el!: HTMLBqOptionElement;
+
+  // State() variables
+  // Inlined decorator, alphabetical order
+  // =======================================
+
+  @State() hasPrefix: boolean = false;
+
+  /** If true, option element is part of a group */
+  @State() isOptionInGroup: boolean = false;
```
endv-bogdanb, line 33: I would replace this `State` with a property `inset` / `indent` defaulted to `false`.

packages/bee-q/src/components/option/bq-option.tsx
```diff
@@ -0,0 +1,178 @@
...
+    this.hasPrefix = hasSlotContent(this.prefixElem, 'prefix');
+  };
+
+  // render() function
+  // Always the last one in the class.
+  // ===================================
+
+  render() {
+    return (
+      <div
```
endv-bogdanb, line 146: Wouldn't be better if you used `button` instead of `div` ?

HamudeHomsi: I didn't think of it as a `button` specifically, but as a list of option items, having the `role="option"` attribute.
But if you think it is more semantic to have a button, then I'll change it.

packages/bee-q/src/components/option-group/bq-option-group.tsx
```diff
@@ -0,0 +1,85 @@
...
+  // Ordered by their natural call order
+  // =====================================
+
+  componentDidLoad() {
+    const slotElements = this.el.shadowRoot
+      .querySelector<HTMLSlotElement>('[part="group"] > slot')
+      .assignedElements({ flatten: true })
+      .filter((elem: HTMLElement) => isHTMLElement(elem, 'bq-option')) as [HTMLBqOptionElement];
+
+    slotElements.forEach((bqOption: HTMLBqOptionElement) => bqOption?.setPaddingToOption());
```
dgonzalezr, line 43: CleanShot 2023-07-20 at 10 17 22

Can you please provide some comments explaining what's the intention here?

HamudeHomsi: Will do! But this will be refactored soon to get simpler than that.
What it does it sets the padding to the inner option within a group
image

dgonzalezr: Maybe there's no need to use JS for that, you can probably add the style from the parent CSS by targeting the shadow part of the children (just thinking out loud here 🤔 )

packages/bee-q/src/components/option/__tests__/bq-option.e2e.ts
dgonzalezr, line 1: @HamudeHomsi is it possible to fix the e2e test so we can also see the progress in Chromatic storybook, please 🙂? Thank you 🙏🏼 

CleanShot 2023-07-24 at 21 32 12@2x

packages/bee-q/src/components/option/bq-option.tsx
```diff
@@ -0,0 +1,170 @@
...
+  @Event() bqOptionFocus: EventEmitter<HTMLBqOptionElement>;
+
+  /** Handler to be called when item is clicked */
+  @Event() bqOptionClick: EventEmitter<HTMLBqOptionElement>;
+
+  /** Handler to be called on enter key press */
+  @Event() bqOptionEnter: EventEmitter<HTMLBqOptionElement>;
+
+  /** Handler to be called on enter key up */
+  @Event() bqOnEnterKeyUp: EventEmitter<HTMLBqOptionElement>;
```
dgonzalezr, line 65: We should be consistent with our naming conventions for custom events:

* bqBlur
* bqFocus
* bqClick
* bqEnter
* bqKeyUp

packages/bee-q/src/components/option/bq-option.tsx
```diff
@@ -0,0 +1,170 @@
...
+  // Component lifecycle events
+  // Ordered by their natural call order
+  // =====================================
+
+  // Listeners
+  // ==============
+
+  @Listen('keydown')
+  onKeyDown(event: KeyboardEvent) {
+    if (event.key === 'Enter') this.bqOptionEnter.emit(this.el);
```
dgonzalezr, line 76:
```suggestion
    if (event.key !== 'Enter') return;
    this.bqOptionEnter.emit(this.el);
```

packages/bee-q/src/components/option/bq-option.tsx
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  @Listen('keyup')
+  onKeyUp(event: KeyboardEvent) {
+    if (event.key === 'Enter') this.bqOnEnterKeyUp.emit(this.el);
```
dgonzalezr, line 81:
```suggestion
    if (event.key !== 'Enter') return;
    this.bqOnEnterKeyUp.emit(this.el);
```

packages/bee-q/src/components/option/_storybook/bq-option.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+};
+export default meta;
+
+type Story = StoryObj;
+
+const Template = (args: Args) => html`
+  <bq-option-list>
+    <bq-option
+      disabled=${args.disabled}
+      selected=${args.selected}
```
dgonzalezr, line 51: Boolean attributes should be bound with a `?` in front: https://lit.dev/docs/v1/lit-html/template-reference/#binding-types

```suggestion
      ?disabled=${args.disabled}
      ?selected=${args.selected}
```

packages/bee-q/src/components/option/bq-option.tsx
```diff
@@ -0,0 +1,170 @@
...
+            label: true,
+            'no-prefix': !this.hasPrefix,
+          }}
+          part="label"
+        >
+          <slot />
+        </span>
+        <span class="bq-option__child suffix" part="suffix">
+          <slot name="suffix" />
+        </span>
```
dgonzalezr, line 166: I just noticed that these 3 elements have the same CSS class name, why? 🤔

dgonzalezr: Also, the prefix and suffix disappear when any of the prop changes in the component

https://github.com/Endava/bee-q/assets/328492/77d6ef98-97b5-4b45-a99b-4694fb0d2561

dgonzalezr: The way you set the prefix and suffix element in the option storybook template is the culprit of the issue above ☝🏼. Avoid to over-engineer stuff, the simplest way will always be the best 😉 

```ts
// bq-options.stories.ts: #line 47

const Template = (args: Args) => html`
  <bq-option-list>
    <bq-option
      ?disabled=${args.disabled}
      ?selected=${args.selected}
      @bqOptionBlur=${args.bqOptionBlur}
      @bqOptionFocus=${args.bqOptionFocus}
      @bqOptionClick=${args.bqOptionClick}
    >
      <bq-icon name="user" slot="prefix"></bq-icon>
      <span>${args.text}</span>
      <bq-icon name="gear" slot="suffix"></bq-icon>
    </bq-option>
  </bq-option-list>
`;
```

packages/bee-q/src/components/option/scss/bq-option.scss
dgonzalezr, line 1: I think we can simplify a lot the option styles if the right markup and classes are used. For example:

```scss
/* -------------------------------------------------------------------------- */
/*                                Option styles                */
/* -------------------------------------------------------------------------- */

@import './bq-option-variables';

:host {
  @apply block;
}

.bq-option {
  @apply flex w-full cursor-pointer items-center text-[length:--bq-option--font-size];
  @apply py-[--bq-option--paddingY] pe-[--bq-option--padding-end] ps-[--bq-option--padding-start];
  @apply rounded-[--bq-option--border-radius] bg-[--bq-option--background] shadow-[shadow:--bq-option--box-shadow];
  @apply border-[length:--bq-option--border-width] border-[color:--bq-option--border-color];
  // Focus
  @apply [&:not(.disabled)]:focus-visible:focus;
  // Hover
  @apply [&:not(.disabled)]:hover:bg-ui-secondary-hover;
  // Active
  @apply [&.active:not(.disabled)]:bg-ui-brand-light [&.active:not(.disabled)]:text-text-brand;
  // Disabled
  @apply [&.disabled]:bg-ui-secondary-disabled [&.disabled]:text-text-primary-disabled;

  border-style: var(--bq-option--border-style);
}

.bq-option--label {
  @apply inline-block flex-auto overflow-hidden text-ellipsis whitespace-nowrap;
  @apply pe-[--bq-option-label--padding-end] ps-[--bq-option-label--padding-start];
}

```

packages/bee-q/src/components/option/bq-option.tsx
dgonzalezr, line 1: Same for the `bq-option` component logic, perhaps something like the following could be a better approach to follow?

```tsx
import { h, Component, Prop, Element, State, Event, EventEmitter, Listen } from '@stencil/core';

import { hasSlotContent } from '../../shared/utils';

/**
 * @part base - The component's internal wrapper.
 * @part label - The `span` element in which the label text is displayed.
 * @part prefix - The `span` element in which the prefix is displayed (generally `bq-icon`).
 * @part suffix - The `span` element in which the suffix is displayed (generally `bq-icon`).
 */
@Component({
  tag: 'bq-option',
  styleUrl: './scss/bq-option.scss',
  shadow: true,
})
export class BqOption {
  // Own Properties
  // ====================

  private prefixElem: HTMLElement;
  private suffixElem: HTMLElement;

  // Reference to host HTML element

  @Element() el!: HTMLBqOptionElement;

  // State() variables
  // Inlined decorator, alphabetical order
  // =======================================

  @State() hasPrefix: boolean = false;
  @State() hasSuffix: boolean = false;

  // Public Property API
  // ========================

  /** If true, the option is disabled. */
  @Prop({ reflect: true }) disabled?: boolean = false;

  /** A string representing the value of the option. Can be used to identify the item */
  @Prop({ reflect: true }) value?: string;

  /** If true, the option is selected and active. */
  @Prop({ reflect: true }) selected: boolean = false;

  // Prop lifecycle events
  // =======================

  // Events section
  // Requires JSDocs for public API documentation
  // ==============================================

  /** Handler to be called when item loses focus */
  @Event() bqBlur: EventEmitter<HTMLBqOptionElement>;

  /** Handler to be called when item is focused */
  @Event() bqFocus: EventEmitter<HTMLBqOptionElement>;

  /** Handler to be called when item is clicked */
  @Event() bqClick: EventEmitter<HTMLBqOptionElement>;

  /** Handler to be called on enter key press */
  @Event() bqEnter: EventEmitter<HTMLBqOptionElement>;

  /** Handler to be called on enter key up */
  @Event() bqkeyUp: EventEmitter<HTMLBqOptionElement>;

  // Component lifecycle events
  // Ordered by their natural call order
  // =====================================

  // Listeners
  // ==============

  @Listen('keydown')
  onKeyDown(event: KeyboardEvent) {
    if (event.key !== 'Enter') return;

    this.bqEnter.emit(this.el);
  }

  @Listen('keyup')
  onKeyUp(event: KeyboardEvent) {
    if (event.key !== 'Enter') return;

    this.bqkeyUp.emit(this.el);
  }

  // Public methods API
  // These methods are exposed on the host element.
  // Always use two lines.
  // Public Methods must be async.
  // Requires JSDocs for public API documentation.
  // ===============================================

  // Local methods
  // Internal business logic.
  // These methods cannot be called from the host element.
  // =======================================================

  private onBlur = (event: Event) => {
    if (this.disabled) {
      event.preventDefault();
      event.stopPropagation();
      return;
    }

    this.bqBlur.emit(this.el);
  };

  private onFocus = (event: Event) => {
    if (this.disabled) {
      event.preventDefault();
      event.stopPropagation();
      return;
    }

    this.bqFocus.emit(this.el);
  };

  private onClick = (event: Event) => {
    if (this.disabled) {
      event.preventDefault();
      event.stopPropagation();
      return;
    }

    this.bqClick.emit(this.el);
  };

  private handlePrefixSlotChange = () => {
    this.hasPrefix = hasSlotContent(this.prefixElem);
  };

  private handleSuffixSlotChange = () => {
    this.hasSuffix = hasSlotContent(this.suffixElem);
  };

  // render() function
  // Always the last one in the class.

  render() {
    return (
      <div
        class={{
          'bq-option': true,
          disabled: this.disabled,
          active: !this.disabled && this.selected,
        }}
        role="option"
        aria-selected={this.selected}
        aria-disabled={this.disabled}
        tabindex={this.disabled ? '-1' : '0'}
        onBlur={this.onBlur}
        onFocus={this.onFocus}
        onClick={this.onClick}
        part="base"
      >
        <span
          class={{
            'bq-option--prefix me-[--bq-option--gap-start] flex items-center': true,
            '!hidden': !this.hasPrefix,
          }}
          ref={(elem) => (this.prefixElem = elem)}
          part="prefix"
        >
          <slot name="prefix" onSlotchange={this.handlePrefixSlotChange} />
        </span>
        <span class="bq-option--label" part="label">
          <slot />
        </span>
        <span
          class={{
            'bq-option--suffix ml-auto ms-[--bq-option--gap-end] flex items-center': true,
            '!hidden': !this.hasSuffix,
          }}
          ref={(elem) => (this.suffixElem = elem)}
          part="suffix"
        >
          <slot name="suffix" onSlotchange={this.handleSuffixSlotChange} />
        </span>
      </div>
    );
  }
}

```

packages/bee-q/src/components/option-list/bq-option-list.tsx
```diff
@@ -0,0 +1,118 @@
...
+  onBqOptionBlur(event: CustomEvent<HTMLElement>) {
+    const { target: item } = event;
+    if (isHTMLElement(item, 'bq-option')) this.bqBlur.emit(item);
+  }
+
+  @Listen('bqOptionFocus', { passive: true })
+  onBqOptionFocus(event: CustomEvent<HTMLElement>) {
+    const { target: item } = event;
+    if (isHTMLElement(item, 'bq-option')) this.bqFocus.emit(item);
+  }
```
dgonzalezr, line 63: I don't think we need to emit blur and focus on the list wrapper, emitting `bqSelect` will be enough for now. I would recommend adding to the `bqSelect` the item value as well:

```tsx
  /** Handler to be called when `bq-option` is selected (on click/enter press). */
  @Event() bqSelect: EventEmitter<{ value: string; item: HTMLBqOptionElement }>;

...

  @Listen('bqClick', { passive: true })
  @Listen('bqEnter', { passive: true })
  onBqSelect(event: CustomEvent<HTMLElement>) {
    const { target: item } = event;

    this.optionItems.forEach((option: HTMLBqOptionElement) => (option.selected = option === item));
    if (isHTMLElement(item, 'bq-option')) {
      this.bqSelect.emit({ item, value: item.value });
    }
  }
```

packages/bee-q/src/components/option-list/bq-option-list.tsx
```diff
@@ -0,0 +1,118 @@
...
+   */
+  private get optionItems(): HTMLBqOptionElement[] | [] {
+    if (!this.baseElem) return [];
+
+    const bqOptionItems: HTMLBqOptionElement[] = [];
+
+    this.slot.assignedElements({ flatten: true }).forEach((elem: HTMLBqOptionElement | HTMLBqOptionGroupElement) => {
+      if (isHTMLElement(elem, 'bq-option')) bqOptionItems.push(elem);
+      if (isHTMLElement(elem, 'bq-option-group'))
+        Array.from(elem.querySelectorAll('bq-option')).forEach((bqOption) => bqOptionItems.push(bqOption));
```
dgonzalezr, line 101: Can you please explain a bit this code block here? What it does do, exactly? Thx!

packages/bee-q/src/components/option-list/bq-option-list.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  // render() function
+  // Always the last one in the class.
+  // ===================================
+
+  render() {
+    return (
+      <div ref={(element) => (this.baseElem = element)} part="base" role="listbox" aria-label="Options">
+        <slot onSlotchange={this.setSlotElem} />
+      </div>
```
dgonzalezr, line 115: According to Figma, there's a gap space of 8px (spacing-xs2) between options. I think you should do the following here:

```tsx
      <div
        class="bq-option--list flex flex-col gap-y-xs"
        ref={(element) => (this.listElem = element)}
        part="base"
        role="listbox"
        aria-label={this.ariaLabel}
      >
        <slot />
      </div>
```

❗ The `aria-label` should not be hardcoded, please.

packages/bee-q/src/components/option/scss/bq-option-variables.scss
```diff
@@ -0,0 +1,41 @@
...
+  --bq-option--font-size: theme('fontSize.m');
+
+  --bq-option--border-color: theme('colors.transparent');
+  --bq-option--border-style: none;
+  --bq-option--border-width: 0;
+  --bq-option--border-radius: theme('borderRadius.s');
+
+  --bq-option--box-shadow: none;
+
+  --bq-option--height: 40px;
```
dgonzalezr, line 35: We can remove this CSS variable, it is not needed 🙂

packages/bee-q/src/components/option/scss/bq-option-variables.scss
```diff
@@ -0,0 +1,41 @@
...
+  --bq-option--padding-start: theme('spacing.s');
+  --bq-option--padding-end: theme('spacing.s');
+  --bq-option--paddingY: theme('spacing.xs');
+  --bq-option-label--padding-start: theme('spacing.s');
+  --bq-option-label--padding-end: theme('spacing.s');
```
dgonzalezr, line 40:
```suggestion
  --bq-option--gap-start: theme('spacing.s');
  --bq-option--gap-end: theme('spacing.s');
  --bq-option--padding-start: theme('spacing.s');
  --bq-option--padding-end: theme('spacing.s');
  --bq-option--paddingY: theme('spacing.xs');
```

Where:

```scss
 /*
  * @prop --bq-option--gap-start: gap space between prefix and label
  * @prop --bq-option--gap-end: gap space between label and suffix
  */ 

```

packages/bee-q/src/components/option-group/bq-option-group.tsx
```diff
@@ -0,0 +1,81 @@
...
+          <span class="option-group__label__child label-text" part="label">
+            <slot name="header-label" />
+          </span>
+          <span class="option-group__label__child suffix" part="suffix">
+            <slot name="header-suffix" />
+          </span>
+        </legend>
+        <div class="option-group__container" role="group" aria-label="Options" part="group">
+          <slot />
+        </div>
```
dgonzalezr, line 77: Shouldn't we follow the same pattern we have in other components for class names? E.g:

```jsx
<Host>
  <legend class="bq-option-group" part="label">
    <span class="bq-option-group--prefix" part="prefix">
      <slot name="header-prefix" />
    </span>
    <span class="bq-option-group--label" part="label">
      <slot name="header-label" />
    </span>
    <span class="bq-option-group--suffix" part="suffix">
      <slot name="header-suffix" />
    </span>
  </legend>
  <div class="bq-option-group--container" role="group" aria-label="Options" part="group">
    <slot />
  </div>
</Host>
```

There's no need of adding an extra CSS class 👇🏼  to be the one to use for targeting the element, instead of

```jsx
<span class="option-group__label__child label-text" part="label">
  <slot name="header-label" />
</span>
```

do

```jsx
<span class="bq-option-group--label" part="label">
  <slot name="header-label" />
</span>
```
```scss
.bq-option-group--label {
  @apply inline-block w-auto overflow-hidden text-ellipsis whitespace-nowrap;
  @apply pe-[--bq-option-group--label-text-padding-end] ps-[--bq-option-group--label-text-padding-start];
}
```

Remember the `K.I.S.S` principle, Keep It Simple Stupid 🙂

packages/bee-q/src/components/option-list/bq-option-list.tsx
dgonzalezr: We can simplify this code by using `querySelectorAll` to get all the `bq-option` elements, no matter if they are inside a `bq-option-group` or not.

```typescript
  // Reference to host HTML element

  @Element() el!: HTMLBqOptionListElement;

  ...

  private get optionItems(): HTMLBqOptionElement[] {
    return Array.from(this.el.querySelectorAll('bq-option'));
  }
```

In this case, `querySelectorAll` will work because the bq-options are not inside the shadow DOM.

packages/bee-q/src/components/option/bq-option.tsx
```diff
@@ -0,0 +1,170 @@
...
+        role="option"
+        aria-selected={this.selected}
+        aria-disabled={this.disabled}
+        tabindex={this.disabled ? '-1' : '0'}
+        onBlur={this.onBlur}
+        onFocus={this.onFocus}
+        onClick={this.onClick}
+        part="base"
+      >
+        <span class="bq-option__child" ref={(elem) => (this.prefixElem = elem)} part="prefix">
```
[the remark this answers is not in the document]
HamudeHomsi: No specific reason, only that these elements (prefix, label and suffix) should have a common styling, so I named a class as `child` as the element description to be generic.
I will change the naming of the class into `bq-option__prefix`, `bq-option__label`, `bq-option__suffix` and apply TW utility classes `flex items-center`

packages/bee-q/src/components/option/bq-option.tsx
HamudeHomsi: Actually here the intention was to have only 1 template for the stories. So, there are 3 stories (`Default`, `Active` and `Disabled`). For `Disabled` story, I wanted to have a different icons displayed in `prefix` and `suffix` slot. So, I needed to configure the template based on `args`. This way we would had only 1 template for all and not duplicate template based on 2 nodes.
I don't have a better solution than to have a separate template for `Disabled` story. (or maybe not to change the icons at all, and then no need for a separate template).

packages/bee-q/src/components/option/scss/bq-option.scss
HamudeHomsi: Great improvement

packages/bee-q/src/components/option/bq-option.tsx
dgonzalezr: You can have a hidden `iconName` argument, and `showIcon`, and based on that you will show the `bq-icon` or not ([see this example here](https://github.com/Endava/bee-q/blob/develop/packages/bee-q/src/components/input/_storybook/bq-input.stories.tsx))... but the current solution is not stable since when any property changes, the element object you're passing down is removed.

packages/bee-q/src/components/option/bq-option.tsx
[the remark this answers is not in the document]
dgonzalezr: You can still do something like:

```css
.bq-optoin--label,
.bq-optoin--prefix,
.bq-optoin--suffix {
  ...
}
```

to apply the same style at once.

dgonzalezr (approved):

PR #266, opened 2023-05-29
feature(breadcrumb): add new `bq-breadcrumb` component

HamudeHomsi:


dgonzalezr:
# ☁️ Nx Cloud Report

> ⚠️ Important: This organization may reach the limit of the FREE Plan and could be disabled at the end of the month. If this happens, remote caching and distributed task execution will no longer work. Your CI may stop working.
        

We didn't find any information for the current pull request with the commit 9ca6c83c1d81f81b9a7d8989eaf24879393e7e9f.
Please verify you are running the latest version of the [NxCloud runner](https://www.npmjs.com/package/nx-cloud).

Check [the Nx Cloud Github Integration documentation](https://nx.dev/nx-cloud/set-up/github) for more information.

Alternatively, you can contact us at [cloud-support@nrwl.io](mailto:cloud-support@nrwl.io).

---

Sent with 💌 from [NxCloud.](https://nx.app)

packages/bee-q/src/components/breadcrumb/bq-breadcrumb.tsx
```diff
@@ -0,0 +1,114 @@
...
+  // =======================================
+
+  // Public Property API
+  // ========================
+
+  // Prop lifecycle events
+  // =======================
+
+  /** The icon name used as separator. Default is `/`. */
+  @Prop() separatorIcon!: string;
```
endv-bogdanb, line 30: Same as above

packages/bee-q/src/components/breadcrumb/bq-breadcrumb.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  componentDidLoad() {
+    this.setIconAsSeparator();
+    this.hideSeparatorFromLastItem();
+  }
+
+  // Listeners
+  // ==============
+
+  @Listen('bqBlur')
```
endv-bogdanb, line 57:
```suggestion
  @Listen('bqBlur', {passive: true})
```

packages/bee-q/src/components/breadcrumb/bq-breadcrumb.tsx
```diff
@@ -0,0 +1,114 @@
...
+  onBlur(event: CustomEvent<HTMLElement>) {
+    if (isHTMLElement(event.detail, 'bq-breadcrumb-item')) this.bqBreadcrumbBlur.emit(event.detail);
+  }
+
+  @Listen('bqFocus')
```
endv-bogdanb, line 62:
```suggestion
  @Listen('bqFocus', {passive: true})
```

packages/bee-q/src/components/breadcrumb/bq-breadcrumb.tsx
```diff
@@ -0,0 +1,114 @@
...
+  onFocus(event: CustomEvent<HTMLElement>) {
+    if (isHTMLElement(event.detail, 'bq-breadcrumb-item')) this.bqBreadcrumbFocus.emit(event.detail);
+  }
+
+  @Listen('bqClick')
```
endv-bogdanb, line 67:
```suggestion
  @Listen('bqClick', {passive: true})
```

packages/bee-q/src/components/breadcrumb/bq-breadcrumb.tsx
```diff
@@ -0,0 +1,114 @@
...
+  /** Handler to be called when `bq-breadcrumb-item` is selected (on click/enter press). */
+  @Event() bqBreadcrumbClick: EventEmitter<HTMLBqBreadcrumbItemElement>;
+
+  // Component lifecycle events
+  // Ordered by their natural call order
+  // =====================================
+
+  componentDidLoad() {
+    this.setIconAsSeparator();
+    this.hideSeparatorFromLastItem();
```
endv-bogdanb, line 51:
```suggestion
```

packages/bee-q/src/components/breadcrumb/bq-breadcrumb.tsx
```diff
@@ -0,0 +1,114 @@
...
+  // =======================================================
+
+  private hideSeparatorFromLastItem = () => {
+    const elements = this.breadcrumbItems;
+    elements.slice(-1)[0].isLast = true;
+  };
+
+  private setIconAsSeparator = (): void => {
+    const lastIndex: number = this.breadcrumbItems.length - 1;
+    this.breadcrumbItems.forEach((item: HTMLBqBreadcrumbItemElement, index: number) =>
```
endv-bogdanb, line 91:
```suggestion
    this.breadcrumbItems.forEach((item, index, arr) =>{
```

packages/bee-q/src/components/breadcrumb/bq-breadcrumb.tsx
```diff
@@ -0,0 +1,114 @@
...
+      index !== lastIndex ? (item.separatorIcon = this.separatorIcon) : undefined,
```
endv-bogdanb, line 92:
```suggestion
      item.separatorIcon = this.separatorIcon;
      item.isLast = index === arr.length - 1;
      }
```

packages/bee-q/src/components/breadcrumb/bq-breadcrumb.tsx
```diff
@@ -0,0 +1,114 @@
...
+    );
+  };
+
+  private get breadcrumbItems(): HTMLBqBreadcrumbItemElement[] {
+    return this.el.shadowRoot
```
endv-bogdanb, line 97:
```suggestion
    return Array.from(this.el.querySelectorAll('bq-breadcrumb-item'))
```

endv-bogdanb:
Rather than using `isLast` I would replace it with `hideSeparatorIcon`, since that's the intend of isLast.

packages/bee-q/src/components/breadcrumb-item/bq-breadcrumb-item.tsx
```diff
@@ -0,0 +1,161 @@
...
+  @State() hasSuffix: boolean = false;
+
+  // Public Property API
+  // ========================
+
+  // Prop lifecycle events
+  // =======================
+
+  /** If true, the item is the last element inside breadcrumb */
+  @Prop() isLast: boolean = false;
```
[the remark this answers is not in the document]
HamudeHomsi: Good catch

packages/bee-q/src/components/breadcrumb/bq-breadcrumb.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  /** Handler to be called when `bq-breadcrumb-item` is selected (on click/enter press). */
+  @Event() bqBreadcrumbClick: EventEmitter<HTMLBqBreadcrumbItemElement>;
+
+  // Component lifecycle events
+  // Ordered by their natural call order
+  // =====================================
+
+  componentDidLoad() {
+    this.setIconAsSeparator();
```
endv-bogdanb, line 50: I would add `@Watch('separatorIcon')`. If they change dynamically the separator we would want to update it automatically.

packages/bee-q/src/components/breadcrumb-item/bq-breadcrumb-item.tsx
```diff
@@ -0,0 +1,153 @@
...
+              'pr-xs': this.hasSuffix && this.hasLabel,
+            }}
+            ref={(elem) => (this.labelElem = elem)}
+            part="label"
+          >
+            <slot name="label" onSlotchange={this.onSlotChange}></slot>
+          </span>
+          <span class="breadcrumb-item__suffix" ref={(elem) => (this.suffixElem = elem)} part="suffix">
+            <slot name="suffix" onSlotchange={this.onSlotChange}></slot>
+          </span>
```
dgonzalezr, line 140: I wonder if we really need 3 slots here 🤔 ... I think it will be easier to have one slot and allows consumers to create the breadcrumb layout as they wish:
- just label
- prefix and label
- label and suffix
- prefix, label, and suffix

Something like:

```html
<bq-breadcrumb-item>
  <div class="flex gap-m">
    <bq-icon name="house"></bq-icon>
    <span>Home</span>
  </div>
</bq-breadcrumb-item>
```

making consumers responsible for applying the CSS rules over the content inside the breadcrumb?

Any thoughts on this @endv-bogdanb ?

HamudeHomsi: I also prefer to keep things simple whenever it's possible. Here, I followed the design from Figma, where we can have items displaying only 1 of the 3 parts, so only icon or only label or all of them, but for this particular component, I think that we can simplify things and allow the consumer to add their preferred spacing.

packages/bee-q/src/components/breadcrumb-item/bq-breadcrumb-item.tsx
```diff
@@ -0,0 +1,133 @@
...
+    const isLink = isDefined(this.href);
+    const TagElem = isLink ? 'a' : 'button';
+    const separatorElem = this.separatorIcon ? (
+      <bq-icon size="12" name={this.separatorIcon}></bq-icon>
+    ) : (
+      <span class="flex h-3 w-3 items-center justify-center">/</span>
+    );
+
+    return (
+      <section class="flex h-5 items-center" role="listitem">
```
dgonzalezr, line 102: Is there a specific reason why we're using `<section>`?

CleanShot 2023-07-10 at 18 08 23

packages/bee-q/src/components/breadcrumb-item/bq-breadcrumb-item.tsx
```diff
@@ -0,0 +1,133 @@
...
+    this.bqClick.emit(this.el);
+  };
+
+  // render() function
+  // Always the last one in the class.
+  // ===================================
+
+  render() {
+    const isLink = isDefined(this.href);
+    const TagElem = isLink ? 'a' : 'button';
```
dgonzalezr, line 94: I don't think we need a `button` for the breadcrumb navigation, the acts as links that navigate users to a page, right?

packages/bee-q/src/components/breadcrumb-item/bq-breadcrumb-item.tsx
```diff
@@ -0,0 +1,133 @@
...
+  // =======================================
+
+  // Public Property API
+  // ========================
+
+  /** If true, the item is the last element inside breadcrumb */
+  @Prop() hideSeparatorIcon: boolean = false;
+
+  /** The icon name used as separator. Default is `/`. */
+  @Prop() separatorIcon!: string;
```
dgonzalezr, line 30: Shouldn't the separator icon be at the parent level (`bq-breadcrumb`)? Otherwise, a consumer could end up with different separators, right? Also, it will be easy (in case the consumer would like to use a custom one) to set it once and not on each breadcrumb item.

packages/bee-q/src/components/breadcrumb-item/bq-breadcrumb-item.tsx
```diff
@@ -0,0 +1,133 @@
...
+        <TagElem
+          class="breadcrumb-item"
+          tabindex="0"
+          href={isLink ? this.href : undefined}
+          rel={isLink && this.target ? 'noreferrer noopener' : undefined}
+          target={isLink ? this.target : undefined}
+          onBlur={this.onBlur}
+          onFocus={this.onFocus}
+          onClick={this.onClick}
+        >
```
dgonzalezr, line 112: the shadow part is missing

packages/bee-q/src/components/breadcrumb/bq-breadcrumb.tsx
```diff
@@ -0,0 +1,106 @@
...
+  /** The icon name used as separator. Default is `/`. */
+  @Prop() separatorIcon!: string;
+
+  // Prop lifecycle events
+  // =======================
+
+  @Watch('separatorIcon')
+  onIconChange() {
+    this.setSeparator();
+  }
```
dgonzalezr, line 35:
```suggestion
  handleSeparatorChange() {
    this.setSeparator();
  }
```

packages/bee-q/src/components/breadcrumb-item/bq-breadcrumb-item.tsx
HamudeHomsi: Don't have a strong argument why I use `<section>` here besides that I usually use this tag as a wrapper in components. Should this be changed to `<div>`?

packages/bee-q/src/components/breadcrumb-item/bq-breadcrumb-item.tsx
HamudeHomsi: Yes, default tag element for the items is `button`. But, if `href` is provided, then we use anchor tag for navigation to a different path or to a different section within the same path.
Maybe we should remove the `button` and only use the `a` element. And if `href` is not provided then use `span`? (maybe the consumer only wants to show a read-only breadcrumb)

packages/bee-q/src/components/breadcrumb-item/bq-breadcrumb-item.tsx
HamudeHomsi: The separator icon is on the parent level, and this variable within the item component it can only be set from the `bq-breadcrumb`. The variable has 2 states:
- undefined => use `/` as separator;
- icon name => display `bq-icon` and `separatorIcon` holds the name set by user

packages/bee-q/src/components/breadcrumb/_storybook/bq-breadcrumb.mdx
```diff
@@ -0,0 +1,35 @@
...
+- Overuse breadcrumbs, or use them in situations where they don't provide any meaningful context or benefit to the user.
+- Use breadcrumbs as a substitute for proper navigation, as they should complement, not replace, other navigation elements such as menus and links.
+- Make breadcrumbs too small or hard to read, as this can make it difficult for users to understand the context and navigate the site effectively.
+- Use breadcrumbs as a way to display too much information, as this can make the navigation overwhelming and confusing for users.
+
+<Title>Properties</Title>
+
+<ArgTypes of="bq-breadcrumb" />
+
+<Stories title="Usage examples" />
```
dgonzalezr, line 35: We can remove Stories and rather add:

```mdx
<ArgTypes of="bq-breadcrumb-item" />
```

packages/bee-q/src/components/breadcrumb/_storybook/bq-breadcrumb.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+  component: 'bq-breadcrumb',
+  parameters: {
+    docs: {
+      page: mdx,
+    },
+  },
+  argTypes: {
+    text: { control: 'text', table: { disable: true } },
+    ['separator-icon']: { control: 'text' },
+  },
```
dgonzalezr, line 18: Please check other components storybook examples and see how events are also added as ArgTypes 😉

packages/bee-q/src/components/breadcrumb/_storybook/bq-breadcrumb.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+        <bq-icon name="house-line" size="16"></bq-icon>
+      </div>
+    </bq-breadcrumb-item>
+    <bq-breadcrumb-item>Men's Clothing</bq-breadcrumb-item>
+    <bq-breadcrumb-item>
+      <div class="flex items-center gap-xs">
+        <bq-icon name="shirt-folded" size="16"></bq-icon>
+        Shirts
+      </div>
+    </bq-breadcrumb-item>
```
dgonzalezr, line 41: Actually, with one single slot, we can still guarantee the layout, my point is to have consumers do something like:

```html
    <bq-breadcrumb-item>
      <bq-icon name="shirt-folded" size="16"></bq-icon>
      Shirts
    </bq-breadcrumb-item>
```

packages/bee-q/src/components/breadcrumb-item/scss/bq-breadcrumb-item.scss
```diff
@@ -0,0 +1,19 @@
+/* -------------------------------------------------------------------------- */
+/*                                Breadcrumb item styles                */
+/* -------------------------------------------------------------------------- */
+
+@import './bq-breadcrumb-item.variables';
+
+.breadcrumb-item {
+  @apply font-regular leading-regular text-text-secondary no-underline focus-visible:focus active:outline-none;
+  @apply px-[var(--bq-breadcrumb-item--paddingX)] py-[var(--bq-breadcrumb-item--paddingY)] text-[length:var(--bq-breadcrumb-item--font-size)];
```
dgonzalezr, line 9: We should use CSS logical properties more instead of padding left/right.
https://tailwindcss.com/docs/padding#using-logical-properties

packages/bee-q/src/components/breadcrumb-item/bq-breadcrumb-item.tsx
```diff
@@ -0,0 +1,133 @@
...
+      <bq-icon size="12" name={this.separatorIcon}></bq-icon>
+    ) : (
+      <span class="flex h-3 w-3 items-center justify-center">/</span>
+    );
+
+    return (
+      <section class="flex h-5 items-center" role="listitem">
+        <TagElem
+          class="breadcrumb-item"
+          tabindex="0"
```
dgonzalezr, line 105: Defining the `tabindex` is unnecessary for link/button

packages/bee-q/src/components/breadcrumb-item/bq-breadcrumb-item.tsx
HamudeHomsi: The `section` is replaced with `Host`

packages/bee-q/src/components/breadcrumb/bq-breadcrumb.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  // render() function
+  // Always the last one in the class.
+  // ===================================
+
+  render() {
+    return (
+      <Host role="list">
+        <slot onSlotchange={this.setSeparator}></slot>
+      </Host>
```
dgonzalezr, line 103: Maybe we can use `<nav>` instead

```suggestion
      <nav role="list">
        <slot onSlotchange={this.setSeparator}></slot>
      </nav>
```

packages/bee-q/src/components/breadcrumb/bq-breadcrumb.tsx
dgonzalezr: @HamudeHomsi can we rename this method, please?

packages/bee-q/src/components/breadcrumb-item/bq-breadcrumb-item.tsx
```diff
@@ -0,0 +1,140 @@
...
+    const isLink = isDefined(this.href);
+    const TagElem = isLink ? 'a' : 'button';
+    const separatorElem = this.separatorIcon ? (
+      <bq-icon size="12" name={this.separatorIcon}></bq-icon>
+    ) : (
+      <span class="flex h-3 w-3 items-center justify-center">/</span>
+    );
+
+    return (
+      <Host class="flex h-5 items-center" role="listitem">
```
dgonzalezr, line 107: Please use a `<div>` instead, there's an issue where Tailwind CSS classes do not get applied to the `<Host>`

dgonzalezr: And do not use hardcoded height (`h-5`), please. The hight should be in accordance with the padding-top/bottom, font-size, and line-height.

packages/bee-q/src/components/breadcrumb/bq-breadcrumb.tsx
```diff
@@ -98,9 +98,9 @@ export class BqBreadcrumb {
 
   render() {
     return (
-      <Host role="list">
+      <nav class="breadcrumb" role="list">
```
dgonzalezr, line 101: You could have left the Tailwind CSS classes, they don't get applied only in the `<Host>` tag.

dgonzalezr: Also a shadow part is missing here 🙂

packages/bee-q/src/components/breadcrumb/scss/bq-breadcrumb.scss
```diff
@@ -2,6 +2,6 @@
 /*                                Breadcrumb styles                */
 /* -------------------------------------------------------------------------- */
 
-:host {
+.breadcrumb {
```
dgonzalezr, line 5:
```css
:host {
  @apply block;
}
```

packages/bee-q/src/components/breadcrumb/bq-breadcrumb.tsx
```diff
@@ -0,0 +1,109 @@
...
+    return Array.from(this.el.querySelectorAll('bq-breadcrumb-item'));
+  }
+
+  // render() function
+  // Always the last one in the class.
+  // ===================================
+
+  render() {
+    return (
+      <nav class="flex items-center" role="list" part="navigation">
```
dgonzalezr, line 104:
```suggestion
      <nav class="flex items-center" part="navigation">
```

packages/bee-q/src/components/breadcrumb/bq-breadcrumb.tsx
```diff
@@ -101,7 +101,7 @@ export class BqBreadcrumb {
 
   render() {
     return (
-      <nav class="flex items-center" role="list" part="navigation">
+      <nav class="flex items-center" role="list" aria-label="breadcrumbs" part="navigation">
```
dgonzalezr: I think the value of this `aria-label` should be a property of `bq-breadcrumb`, what do you think?

packages/bee-q/src/components/breadcrumb/bq-breadcrumb.tsx
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  // render() function
+  // Always the last one in the class.
+  // ===================================
+
+  render() {
+    return (
+      <Host>
+        <nav class="flex items-center" role="list" aria-label="breadcrumbs" part="navigation">
```
dgonzalezr, line 125: I think the value of this `aria-label` should be a property of `bq-breadcrumb`, what do you think?

packages/bee-q/src/components/breadcrumb/bq-breadcrumb.tsx
```diff
@@ -0,0 +1,135 @@
...
+      .assignedElements({ flatten: true })[0] as HTMLElement;
+  }
+
+  private get defaultSeparator(): HTMLElement {
+    const newNode = document.createElement('span');
+    newNode.classList.add('flex', 'w-3', 'items-center', 'justify-center');
+    newNode.textContent = '/';
+
+    return newNode;
+  }
```
dgonzalezr, line 116: I just tested adding a custom separator and this is what happened:

CleanShot 2023-07-12 at 20 02 29

Also, if the separator change, it is not reflected, eg. if I change the custom separator icon size or the name.

One more thing, it won't be easier if we just append the separator after each breadcrumb item (except the last one) instead of adding it inside a `bq-breadcrumb-item` slot?

dgonzalezr, line 116: You can add a `<bq-icon>` inside the slot separator and then the logic to clone it as you do for a custom one. That way you don't need to worry if a custom separator is set, to remove the default, it will be overridden by the external slot content.

packages/bee-q/src/components/breadcrumb/bq-breadcrumb.tsx
HamudeHomsi: Agree!

packages/bee-q/src/components/breadcrumb/bq-breadcrumb.tsx
HamudeHomsi: I see your point. From integration it works as intended, but in case of changing the separator dynamically or manually it doesn't work. I'll fix it.
I think that the separator element is relevant in the same context as breadcrumb item. Let s say we remove an item, then we want to remove the inner text and also the separator, together. That's why the separator is placed inside the item. Thoughts?

dgonzalezr: Yes, it makes sense regarding where to attach the separator. Let's keep it that way then 👍🏼

HamudeHomsi: Nice. Will change it!

packages/bee-q/src/components/breadcrumb-item/bq-breadcrumb-item.tsx
```diff
@@ -0,0 +1,132 @@
...
+        <TagElem
+          class="breadcrumb-item"
+          href={isLink ? this.href : undefined}
+          rel={isLink && this.target ? 'noreferrer noopener' : undefined}
+          target={isLink ? this.target : undefined}
+          onBlur={this.onBlur}
+          onFocus={this.onFocus}
+          onClick={this.onClick}
+          aria-current={isLink && this.isLastItem ? 'location' : undefined} // indicates the last link as the current page
+          part="base"
```
dgonzalezr, line 109: `part=base` should be in the parent div while here it could be: `part="item"`

packages/bee-q/src/components/breadcrumb/bq-breadcrumb.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  // State() variables
+  // Inlined decorator, alphabetical order
+  // =======================================
+
+  // Public Property API
+  // ========================
+
+  /** The `aria-label` attribute for `<nav>` */
+  @Prop({ reflect: true }) ariaLabel: string = 'breadcrumbs';
```
dgonzalezr, line 32:
```suggestion
  /** The `aria-label` attribute to describe the type of navigation */
  @Prop({ reflect: true }) ariaLabel: string;
```

packages/bee-q/src/components/breadcrumb-item/bq-breadcrumb-item.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+  // State() variables
+  // Inlined decorator, alphabetical order
+  // =======================================
+
+  // Public Property API
+  // ========================
+
+  /** If true, the item is the last element inside breadcrumb */
+  @Prop() isLastItem: boolean = false;
```
dgonzalezr, line 33: We should add an `aria-label` for the breadcrumb item that will allow consumers to place the page title in context for screen readers.

```suggestion
  /** The aria-label that corresponds to the full title of the destination page.
   * This won't be shown in the page, but it will be used by screen readers and other assistive devices */
  @Prop() ariaLabel: string;

  /** If true, the item is the last element inside breadcrumb */
  @Prop() isLastItem: boolean = false;
```

Usage example:

```html
<bq-breadcrumb-item aria-label="Hom page">
  <bq-icon name="house-line" size="16"></bq-icon>
</bq-breadcrumb-item>
```

packages/bee-q/src/components/breadcrumb-item/bq-breadcrumb-item.tsx
```diff
@@ -0,0 +1,133 @@
...
+  // render() function
+  // Always the last one in the class.
+  // ===================================
+
+  render() {
+    const isLink = isDefined(this.href);
+    const TagElem = isLink ? 'a' : 'button';
+
+    return (
+      <div class="flex items-center" part="base">
```
dgonzalezr, line 100:
```suggestion
      <div class="flex items-center" aria-label={this.ariaLabel} part="base">
```

packages/bee-q/src/components/breadcrumb/bq-breadcrumb.tsx
HamudeHomsi: Shouldn't we use a default `aria-label` value for the `nav`? In case the user will not add any value then we have `breadcrumb` as default.

dgonzalezr: Then use `Breadcrumbs` (with Capitalize letter) please 🙂

packages/bee-q/src/components/breadcrumb/_storybook/bq-breadcrumb.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+  component: 'bq-breadcrumb',
+  parameters: {
+    docs: {
+      page: mdx,
+    },
+  },
+  argTypes: {
+    text: { control: 'text', table: { disable: true } },
+    ['aria-label']: { control: 'text' },
+    htmlNode: { control: { type: 'object' } },
```
dgonzalezr, line 18: We don't want this to be shown in the control panel, since it is not part of the component

```suggestion
    htmlNode: { control: 'object',  table: { disable: true } },
```

packages/bee-q/src/components/breadcrumb/_storybook/bq-breadcrumb.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+};
+
+export const Links: Story = {
+  render: (args: Args) => html`
+    <bq-breadcrumb
+      @bqBreadcrumbBlur=${args.bqBreadcrumbBlur}
+      @bqBreadcrumbClick=${args.bqBreadcrumbClick}
+      @bqBreadcrumbFocus=${args.bqBreadcrumbFocus}
+    >
+      <bq-breadcrumb-item href="https://example.com/" target="_blank" aria-label="Home page">
```
dgonzalezr, line 72: No need to created another `html``` output, you could have a hidden boolean argument, eg: `useLinks` that when true it will output the href link value, something like:

```tsx
<bq-breadcrumb-item ${ args.useLinks ? 'href="https://example.com/" target="_blank"' : ''} aria-label="Home page">
```

packages/bee-q/src/components/breadcrumb/_storybook/bq-breadcrumb.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+  argTypes: {
+    text: { control: 'text', table: { disable: true } },
+    ['aria-label']: { control: 'text' },
+    htmlNode: { control: 'object', table: { disable: true } },
+    useLinks: { control: 'boolean', table: { disable: true } },
+    // Event handlers
+    bqBreadcrumbBlur: { action: 'bqBlur' },
+    bqBreadcrumbClick: { action: 'bqClick' },
+    bqBreadcrumbFocus: { action: 'bqFocus' },
+  },
```
dgonzalezr, line 26:
```suggestion
  argTypes: {
    'aria-label': { control: 'text' },
    // Event handlers
    bqBreadcrumbBlur: { action: 'bqBlur' },
    bqBreadcrumbClick: { action: 'bqClick' },
    bqBreadcrumbFocus: { action: 'bqFocus' },
    // Not part of the public API, so we don't want to expose it in the docs
    text: { control: 'text', table: { disable: true } },
    htmlNode: { control: 'object', table: { disable: true } },
    useLinks: { control: 'boolean', table: { disable: true } },
  },
```

dgonzalezr (approved):

PR #694, opened 2023-12-06
feat(Empty State): add new bq-empty-state component

Cata1989:
An Empty State is a UI component that is displayed when a user interacts with an application or system and there is no data or content available to display. Empty States are common in applications that have dynamic or changing data, or when the user is in a state of no activity or interaction. 

https://github.com/Endava/BEEQ/assets/98550729/0f3e03cf-96bc-4269-a5b0-2d16795489e0


packages/beeq/src/components/empty-state/bq-empty-state.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  // render() function
+  // Always the last one in the class.
+  // ===================================
+
+  render() {
+    return (
+      <Host>
+        <div class="bq-empty-state" part="wrapper">
+          <div class={{ [`bq-empty-state-icon-margin-bottom__${this.size}`]: true }}>
```
dgonzalezr, line 85: Perhaps, we could use a simple pattern like:
```suggestion
          <div class={{ [`bq-empty-state__thumbnail mb__${this.size}`]: true }}>
```

packages/beeq/src/components/empty-state/bq-empty-state.tsx
```diff
@@ -0,0 +1,117 @@
...
+            <slot name="icon">
```
dgonzalezr, line 86:
```suggestion
            <slot name="thumbnail">
```

packages/beeq/src/components/empty-state/bq-empty-state.tsx
```diff
@@ -0,0 +1,117 @@
...
+            </slot>
+          </div>
+          <div
+            class={{
+              'title-font font-bold leading-regular text-text-primary': true,
+              [`bq-empty-state-title-font-size__${this.size}`]: true,
+              [`bq-empty-state-title-margin-bottom__${this.size}`]: true,
+            }}
+            part="title"
+          >
```
dgonzalezr, line 97: I would avoid declaring CSS vars for `font-size` and `margin-bottom`, since consumers can override this easily on integration by wrapping the content inside a div or span container.

```suggestion
          <div
            class={{
              'title-font font-bold leading-regular text-text-primary': true,
              'text-s mb-m': this.size === 'small'
              'text-m mb-l': this.size === 'medium'
              'text-l mb-xl': this.size === 'large'
            }}
            part="title"
          >
```

packages/beeq/src/components/empty-state/bq-empty-state.tsx
```diff
@@ -0,0 +1,117 @@
...
+            <slot />
+          </div>
+          <div
+            class={{
+              'font-normal leading-regular': true,
+              [`bq-empty-state-body-font-size__${this.size}`]: true,
+              [`bq-empty-state-body-margin-bottom__${this.size}`]: true,
+            }}
+            part="body"
+          >
```
dgonzalezr, line 107: Same here

```suggestion
          <div
            class={{
              'font-normal leading-regular': true,
              'mb-m text-s': this.size === 'small',
              'mb-l text-m': this.size === 'medium',
              'mb-xl text-l': this.size === 'large',
            }}
            part="body"
          >
```

packages/beeq/src/components/empty-state/bq-empty-state.tsx
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  // render() function
+  // Always the last one in the class.
+  // ===================================
+
+  render() {
+    return (
+      <Host>
+        <div class="bq-empty-state" part="wrapper">
```
dgonzalezr, line 84: You can move here the TW utility classes from the .scss file:

```suggestion
      <div class="bq-empty-state flex flex-col items-center" part="wrapper">
```

packages/beeq/src/components/empty-state/bq-empty-state.tsx
```diff
@@ -0,0 +1,117 @@
...
+    return SIZE_TO_VALUE_MAP[this.size] || SIZE_TO_VALUE_MAP.medium;
+  }
+
+  // render() function
+  // Always the last one in the class.
+  // ===================================
+
+  render() {
+    return (
+      <Host>
```
dgonzalezr, line 83: `Host` can be removed, since the div wrapper already group all the elements and there's nothing applied to the `<Host>`

packages/beeq/src/components/empty-state/bq-empty-state.tsx
dgonzalezr: Use `inline-flex` instead of `flex`

packages/beeq/src/components/empty-state/bq-empty-state.tsx
dgonzalezr: Keep in mind that margin-bottom should only applied if there's body content, otherwise the component takes  more space than it should

CleanShot 2023-12-06 at 16 50 33

packages/beeq/src/components/empty-state/scss/bq-empty-state.variables.scss
dgonzalezr, line 1: Considering that Empty state is super simple and flexible, I think we can remove these CSS vars

packages/beeq/src/components/empty-state/bq-empty-state.tsx
dgonzalezr: Notice that we're missing the shadow DOM part definition here.

```
part="thumbnail"
```

packages/beeq/src/components/empty-state/_storybook/bq-empty-state.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+export const WithBody: Story = {
+  render: TemplateWithBody,
+  args: {},
+};
+
+export const WithCallToAction: Story = {
+  render: TemplateWithCTA,
+  args: {},
+};
```
dgonzalezr, line 77: If all three templates are completely different, you can move them into the corresponding story. E.g:

```ts
export const Default: Story = {
  render: (args: Args) => html`
    <bq-empty-state size=${args.size}> Title </bq-empty-state>
  `,
};

export const WithBody: Story = {
  render: (args: Args) => html`
    <div class="flex flex-row gap-20">
      <bq-empty-state size=${args.size}>
        Title
        <span slot="body"> Description </span>
      </bq-empty-state>
      <bq-empty-state size=${args.size}>
        Title <span slot="body"> Description <a class="bq-link" href="https://example.com">Link</a> </span>
      </bq-empty-state>
    </div>
  `;
};

...
```

`args` can be removed if you are not assigning new values to arguments.

dgonzalezr (approved):

PR #695, opened 2023-12-07
feat(Tag): add new `bq-tag` component

Cata1989:
The Tag Component is a UI element used to label and categorize content within an application. Tags are commonly used to label items with keywords or categories, making it easier to find and organize content.

Screenshot 2023-12-07 at 12 33 08

Screenshot 2023-12-13 at 14 49 34


packages/beeq/src/components/tag/bq-tag.tsx
```diff
@@ -0,0 +1,221 @@
...
+            active: !this.disabled && this.selected,
+            [`bq-tag__wrapper--${this.size} font-medium leading-regular`]: true,
+            [`bq-tag__${this.type}__${this.variant}`]: this.hasColor,
+          }}
+          aria-selected={this.selected}
+          aria-disabled={this.disabled}
+          onClick={this.handleClick}
+          onFocus={this.handleFocus}
+          onKeyDown={this.handleOnKeyDown}
+          role="tab"
```
dgonzalezr, line 188: Are we sure about this role attribute value for accessibility? I see we have some issues there:

CleanShot 2023-12-14 at 14 11 09

packages/beeq/src/components/tag/bq-tag.tsx
```diff
@@ -0,0 +1,221 @@
...
+          part="wrapper"
+          tabindex={this.disabled ? '-1' : '0'}
+        >
+          <div class={{ 'bq-tag__icon': true, '!hidden': !this.hasIcon }}>
+            <slot name="icon">
+              <bq-icon size={this.iconSize} name="star" part="icon" exportparts="base,svg" />
+            </slot>
```
dgonzalezr, line 195: This should be an empty slot, probably called `prefix`, what do you think?

packages/beeq/src/components/tag/bq-tag.tsx
```diff
@@ -0,0 +1,221 @@
...
+          aria-selected={this.selected}
+          aria-disabled={this.disabled}
+          onClick={this.handleClick}
+          onFocus={this.handleFocus}
+          onKeyDown={this.handleOnKeyDown}
+          role="tab"
+          part="wrapper"
+          tabindex={this.disabled ? '-1' : '0'}
+        >
+          <div class={{ 'bq-tag__icon': true, '!hidden': !this.hasIcon }}>
```
dgonzalezr, line 192: I would recommend using a state instead of a prop, eg:. `this.hasPrefix`.
Then you could use the `hasSlotContent` utility to hide the div container if there's no prefix content.

packages/beeq/src/components/tag/bq-tag.tsx
```diff
@@ -0,0 +1,221 @@
...
+            </slot>
+          </div>
+          <div
+            class={{
+              'text-xs': this.size === 'extra_small',
+              'text-s': this.size === 'small',
+              'text-m': this.size === 'medium',
+            }}
+          >
+            <slot name="tag" />
```
dgonzalezr, line 204: This should be the default `slot` (no name needed), meaning that when using the component see can only use it as:
```html
<bq-tag>text</bq-tag>
```

packages/beeq/src/components/tag/bq-tag.tsx
```diff
@@ -0,0 +1,221 @@
...
+              <bq-icon size={this.iconSize} name="star" part="icon" exportparts="base,svg" />
+            </slot>
+          </div>
+          <div
+            class={{
+              'text-xs': this.size === 'extra_small',
+              'text-s': this.size === 'small',
+              'text-m': this.size === 'medium',
+            }}
+          >
```
dgonzalezr, line 203: We are missing exporting the shadow DOM part here:

```suggestion
          <div
            class={{
              'text-xs': this.size === 'extra_small',
              'text-s': this.size === 'small',
              'text-m': this.size === 'medium',
            }}
            part="text"
          >
```

packages/beeq/src/components/tag/scss/bq-tag.scss
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  &__filled {
+    @apply bg-[color:var(--bq-ui--brand)] text-[color:var(--bq-text--primary-alt)];
+  }
+}
+
+.bq-tag__icon {
+  @apply inline-flex;
+}
```
dgonzalezr, line 101: This can be moved to JSX

packages/beeq/src/components/tag/bq-tag.tsx
```diff
@@ -0,0 +1,221 @@
...
+        <div
+          class={{
+            'bq-tag': true,
+            disabled: this.disabled,
+            active: !this.disabled && this.selected,
+            [`bq-tag__wrapper--${this.size} font-medium leading-regular`]: true,
+            [`bq-tag__${this.type}__${this.variant}`]: this.hasColor,
+          }}
+          aria-selected={this.selected}
+          aria-disabled={this.disabled}
```
dgonzalezr, line 184: This aria-* attributes can be removed, there's no scope for them in this component 😉

packages/beeq/src/components/tag/bq-tag.tsx
dgonzalezr: Yep, make sense to me 👍🏼

packages/beeq/src/components/tag/bq-tag.tsx
```diff
@@ -0,0 +1,233 @@
...
+            [`bq-tag__wrapper--${this.size} font-medium leading-regular`]: true,
+            [`bq-tag__${this.type}__${this.variant}`]: this.hasColor,
+          }}
+          onClick={this.handleClick}
+          onFocus={this.handleFocus}
+          onKeyDown={this.handleOnKeyDown}
+          role="button"
+          part="wrapper"
+          tabindex={this.disabled ? '-1' : '0'}
+        >
```
dgonzalezr, line 200: I think that by changing to `<button>` instead of `<div>`, we can remove the role attribute. The only think is to remove borders on the `.bq-tag` CSS class:

```suggestion
        <button
          class={{
            'bq-tag': true,
            disabled: this.disabled,
            active: !this.disabled && this.selected,
            [`bq-tag__wrapper--${this.size} font-medium leading-regular`]: true,
            [`bq-tag__${this.type}__${this.variant}`]: this.hasColor,
          }}
          onClick={this.handleClick}
          onFocus={this.handleFocus}
          onKeyDown={this.handleOnKeyDown}
          part="wrapper"
          tabindex={this.disabled ? '-1' : '0'}
        >
```

```css
.bq-tag {
  @apply inline-flex border-0 cursor-pointer flex-row items-center justify-center bg-ui-secondary;
```

packages/beeq/src/components/tag/scss/bq-tag.scss
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  &--small {
+    @apply gap-xs2 rounded-xs px-xs py-xs3;
+  }
+
+  &--medium {
+    @apply gap-xs rounded-s px-s py-xs2;
+  }
+}
```
dgonzalezr, line 43: We can move all these to JSX, something like:

```tsx
        <button
          class={{
            'bq-tag gap-xs rounded-s px-s py-xs2 font-medium leading-regular': true,
            active: !this.disabled && this.selected,
            'gap-xs2 rounded-xs px-xs py-xs3': this.size !== 'medium',
            [`bq-tag__${this.type}__${this.variant}`]: this.hasColor,
          }}
```

it might work

packages/beeq/src/components/tag/bq-tag.tsx
dgonzalezr, line 200: If we change to `<button>`, the disabled state should be set through the attribute instead.

```tsx
        <button
          class={{ ... }}
          disabled={this.disabled}
          onClick={this.handleClick}
          onFocus={this.handleFocus}
          onKeyDown={this.handleOnKeyDown}
          part="wrapper"
        >
```
and we can get rid of tabIndex

packages/beeq/src/components/tag/bq-tag.tsx
```diff
@@ -0,0 +1,233 @@
...
+            <slot />
+          </div>
+          {this.isRemovable && !this.hasColor && !this.disabled && (
+            <bq-button
+              class="bq-tag__close"
+              appearance="text"
+              size="small"
+              onClick={() => this.hide()}
+              part="btn-close"
+            >
```
dgonzalezr, line 225: We can use directly `this.handleHide` in the onClick event handler

```suggestion
            <bq-button class="bq-tag__close" appearance="text" size="small" onClick={this.handleHide} part="btn-close">
```

packages/beeq/src/components/tag/_storybook/bq-tag.stories.tsx
```diff
@@ -0,0 +1,294 @@
...
+    open: { control: 'boolean' },
+    type: { control: 'select', options: [...TAG_TYPE] },
+    variant: { control: 'select', options: [...TAG_VARIANT] },
+    disabled: { control: 'boolean' },
+    selected: { control: 'boolean' },
+    'is-removable': { control: 'boolean' },
+    'has-color': { control: 'boolean' },
+    // Event handlers
+    bqClick: { action: 'bqClick' },
+    bqFocus: { action: 'bqFocus' },
```
dgonzalezr, line 26: We are missing other events here

packages/beeq/src/components/tag/_storybook/bq-tag.stories.tsx
```diff
@@ -0,0 +1,294 @@
...
+      </bq-tag>
+    </div>
+  `,
+  args: {
+    open: true,
+    'is-removable': true,
+  },
+};
+
+export const Color: Story = {
```
dgonzalezr, line 165: I will split these into 2 stories: filled, outline

packages/beeq/src/components/tag/bq-tag.tsx
```diff
@@ -0,0 +1,233 @@
...
+  // ========================
+
+  /** The type of the tag component */
+  @Prop({ reflect: true }) size: TTagSize = 'medium';
+
+  /** If true, the tag component has color style */
+  @Prop({ reflect: true }) hasColor: boolean;
+
+  /** If true, the tag component can be removed */
+  @Prop({ reflect: true }) isRemovable: boolean;
```
dgonzalezr, line 43: Can we use just `removable` instead? I think it makes it simpler, what do you think?

packages/beeq/src/components/tag/bq-tag.tsx
```diff
@@ -0,0 +1,233 @@
...
+      event.preventDefault();
+      event.stopPropagation();
+      return;
+    }
+
+    const ev = this.bqClick.emit(this.el);
+    if (!ev.defaultPrevented) {
+      this.selected = !this.selected;
+    }
+  };
```
dgonzalezr, line 151: I think that if the tag is `removable` we should avoid triggering the click event handler.

```suggestion
  private handleClick = (event: Event) => {
    if (this.removable) return;
    
    const ev = this.bqClick.emit(this.el);
    if (!ev.defaultPrevented) {
      this.selected = !this.selected;
    }
  };
```

packages/beeq/src/components/tag/bq-tag.tsx
Cata1989: Yes, sure thing.

packages/beeq/src/components/tag/__tests__/bq-tag.e2e.ts
dgonzalezr, line 1: We definitely need more e2e tests here 🙂

Cata1989: Ohh, yes, I'll continue with adding more tests here. ;)

packages/beeq/src/components/tag/bq-tag.types.ts
```diff
@@ -12,3 +12,30 @@ export const SIZE_TO_VALUE_MAP: Record<TTagSize, number> = {
...
+    filled: 'text--primary-alt',
+  },
+  info: {
+    default: 'text--brand',
+    filled: 'text--primary-alt',
+  },
+} as const;
+
+// the function that returns the color based on the type and variant
+export const getTagColor = (type: TTagType, variant: TTagVariant): string => tagColors[type][variant];
```
dgonzalezr, line 41: This file is intended to be only for types, maybe you would like to move this to the component or create a `helper/index.ts` file instead (at the component level)

Cata1989: I have added this private get method inside the component which is similar with the get iconSize() method.

packages/beeq/src/components/tag/bq-tag.types.ts
```diff
@@ -0,0 +1,38 @@
+export const TAG_SIZE = ['extra_small', 'small', 'medium'] as const;
+export type TTagSize = (typeof TAG_SIZE)[number];
+
+export const TAG_TYPE = ['default', 'success', 'warning', 'error', 'info'] as const;
+export type TTagType = (typeof TAG_TYPE)[number];
+
+export const TAG_VARIANT = ['default', 'filled'];
```
dgonzalezr, line 7:
```suggestion
export const TAG_VARIANT = ['default', 'filled'] as const;
```

packages/beeq/src/components/tag/bq-tag.types.ts
```diff
@@ -0,0 +1,38 @@
...
+export type TTagVariant = (typeof TAG_VARIANT)[number];
+
+export const SIZE_TO_VALUE_MAP: Record<TTagSize, number> = {
+  extra_small: 20,
+  small: 20,
+  medium: 24,
+};
+
+// the definition of colors associated with each type and variant of the label
+export const tagColors = {
```
dgonzalezr, line 17: Use upper case for constants:

```suggestion
export const TAG_COLORS = {
```

packages/beeq/src/components/tag/bq-tag.tsx
```diff
@@ -0,0 +1,230 @@
...
+  @Prop({ reflect: true }) variant: TTagVariant = 'filled';
+
+  /** If true, the Tag will be disabled (only if clickable = `true`, no interaction allowed) */
+  @Prop({ reflect: true }) disabled?: boolean = false;
+
+  /** If true, the Tag can be clickable */
+  @Prop({ reflect: true }) clickable: boolean = false;
+
+  /** If true, the Tag is selected (only if clickable = `true`) */
+  @Prop({ reflect: true, mutable: true }) selected: boolean = false;
```
dgonzalezr, line 61: It will be great if we could sort these props in alphabetical order 🙂

packages/beeq/src/components/tag/bq-tag.tsx
```diff
@@ -0,0 +1,230 @@
...
+  @Event() bqHide: EventEmitter;
+
+  /** Callback handler to be called when the tag is not removable */
+  @Event() bqShow: EventEmitter;
+
+  /** Handler to be called when tag is clicked */
+  @Event() bqClick: EventEmitter<HTMLBqTagElement>;
+
+  /** Handler to be called when tag is focused */
+  @Event() bqFocus: EventEmitter<HTMLBqTagElement>;
```
dgonzalezr, line 87: We should think about emitting a bqClose and bqOpen events for when the tag is hidden/shown

dgonzalezr (approved):
Looking good 🎉

PR #741, opened 2024-01-02
fix(Accordion): skip focus if `disabled="true"`

dgonzalezr:


packages/beeq/src/components/accordion-group/bq-accordion-group.tsx
```diff
@@ -90,7 +90,7 @@ export class BqAccordionGroup {
 
   render() {
     return (
-      <div class="bq-accordion-group">
+      <div class="bq-accordion-group flex flex-col gap-[--bq-accordion-group--gap]">
```
endv-bogdanb, line 93:
```suggestion
      <div class="flex flex-col gap-[--bq-accordion-group--gap]">
```

dgonzalezr: Solved!

endv-bogdanb (approved):

PR #769, opened 2024-01-12
fix(Alert): improve open/close animation

Cata1989:
The purpose of this PR is to improve the open/close animation for Alert component.

### DIFF
--- a/packages/beeq/src/components/alert/__tests__/bq-alert.e2e.ts
+++ b/packages/beeq/src/components/alert/__tests__/bq-alert.e2e.ts
@@ -26,7 +26,7 @@ describe('bq-alert', () => {
 
     const element = await page.find('bq-alert');
     expect(element).toEqualAttribute('aria-hidden', 'true');
-    expect(element).not.toHaveClass('is-hidden');
+    expect(element).toHaveClass('is-hidden');
   });
 
   it('should render as hidden with `open="false"`', async () => {
@@ -36,7 +36,7 @@ describe('bq-alert', () => {
 
     const element = await page.find('bq-alert');
     expect(element).toEqualAttribute('aria-hidden', 'true');
-    expect(element).not.toHaveClass('is-hidden');
+    expect(element).toHaveClass('is-hidden');
   });
 
   it('should render as open', async () => {
--- a/packages/beeq/src/components/alert/__tests__/bq-alert.e2e.ts
+++ b/packages/beeq/src/components/alert/__tests__/bq-alert.e2e.ts
@@ -26,7 +26,7 @@ describe('bq-alert', () => {
 
     const element = await page.find('bq-alert');
     expect(element).toEqualAttribute('aria-hidden', 'true');
-    expect(element).toHaveClass('is-hidden');
+    expect(element).not.toHaveClass('is-hidden');
   });
 
   it('should render as hidden with `open="false"`', async () => {
@@ -36,7 +36,7 @@ describe('bq-alert', () => {
 
     const element = await page.find('bq-alert');
     expect(element).toEqualAttribute('aria-hidden', 'true');
-    expect(element).toHaveClass('is-hidden');
+    expect(element).not.toHaveClass('is-hidden');
   });
 
   it('should render as open', async () => {
--- a/packages/beeq/src/components/alert/bq-alert.tsx
+++ b/packages/beeq/src/components/alert/bq-alert.tsx
@@ -100,11 +100,17 @@ export class BqAlert {
   // Requires JSDocs for public API documentation
   // ==============================================
 
-  /** Callback handler to be called when the notification is hidden */
-  @Event() bqHide: EventEmitter;
+  /** Callback handler to be called when the alert is hidden */
+  @Event() bqHide!: EventEmitter;
 
-  /** Callback handler to be called when the notification is shown */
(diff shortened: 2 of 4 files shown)


packages/beeq/src/components/alert/bq-alert.tsx
```diff
@@ -190,6 +211,13 @@ export class BqAlert {
             [`bq-alert bq-alert__${this.type}`]: true,
             'is-sticky': this.sticky,
           }}
+          data-transition-enter="transform transition ease-out duration-300"
+          data-transition-enter-start="translate-y-xs opacity-0 sm:translate-y-0 sm:translate-x-s"
+          data-transition-enter-end="translate-y-0 opacity-100 sm:translate-x-0"
+          data-transition-leave="transform transition ease-in duration-100"
+          data-transition-leave-start="translate-y-0 opacity-100 sm:translate-x-0"
+          data-transition-leave-end="-translate-y-xs opacity-0 sm:translate-y-0 sm:translate-x-s"
```
dgonzalezr, line 219: I would rather use an opacity transition for the enter/leave animation. Remember that Alert are intended to appear right were they are placed (except for the sticky version)

```suggestion
          data-transition-enter="transition ease-out duration-300"
          data-transition-enter-start="opacity-0"
          data-transition-enter-end="opacity-100"
          data-transition-leave="transition ease-in duration-200"
          data-transition-leave-start="opacity-100"
          data-transition-leave-end="opacity-0"
```

dgonzalezr: Also, we might need to change the `Watch('open')` method a bit.

```ts
  @Watch('open')
  handleOpenChange() {
    this.autoDismissDebounce?.cancel();

    if (!this.open) {
      this.handleHide();
      return;
    }

    this.handleShow();

    if (this.autoDismiss) {
      this.autoDismissDebounce();
    }
  }
```

With the current method, the `handleShow()` and `handleHide()` methods are not hit when the open property values change, meaning no animation is shown, but the change proposed should fix that.


https://github.com/Endava/BEEQ/assets/328492/769484e9-6b04-4d68-b71e-d6722f17eed3

packages/beeq/src/components/alert/scss/bq-alert.scss
dgonzalezr, line 1: Let's also change the `:host(.is-hidden)` rule, so the leave animation can be properly seen:

```scss
:host(.is-hidden) {
  @apply invisible scale-0;
}
```

packages/beeq/src/components/alert/bq-alert.tsx
```diff
@@ -125,31 +133,44 @@ export class BqAlert {
...
 
-  private handleHide = () => {
+  private handleHide = async () => {
     const ev = this.bqHide.emit(this.el);
     if (!ev.defaultPrevented) {
+      await leave(this.alertElement);
       this.open = false;
+      this.handleTransitionEnd();
     }
   };
```
dgonzalezr, line 156: Let's apply the `.is-hidden` class in the Host via JS, right after the leave animation, that way we guarantee that the animation runs properly.

```suggestion
  private handleHide = async () => {
    const ev = this.bqHide.emit(this.el);
    if (!ev.defaultPrevented) {
      await leave(this.alertElement);
      this.el.classList.add('is-hidden');
      this.handleTransitionEnd();
      this.open = false;
    }
  };
```

packages/beeq/src/components/alert/bq-alert.tsx
```diff
@@ -125,31 +133,44 @@ export class BqAlert {
...
 
-  private handleShow = () => {
+  private handleShow = async () => {
     const ev = this.bqShow.emit(this.el);
     if (!ev.defaultPrevented) {
       this.open = true;
+      await enter(this.alertElement);
+      this.handleTransitionEnd();
+    }
+  };
```
dgonzalezr, line 165: Same for here, we remove the `.is-hidden` class before the enter animation

```suggestion
  private handleShow = async () => {
    const ev = this.bqShow.emit(this.el);
    if (!ev.defaultPrevented) {
      this.open = true;
      this.el.classList.remove('is-hidden');
      await enter(this.alertElement);
      this.handleTransitionEnd();
    }
  };
```

dgonzalezr: and with this, we can remove from the `<Host>` the conditional to apply the `.is-hidden` class:

CleanShot 2024-01-12 at 19 05 04@2x

packages/beeq/src/components/alert/__tests__/bq-alert.e2e.ts
```diff
@@ -36,7 +36,7 @@ describe('bq-alert', () => {
 
     const element = await page.find('bq-alert');
     expect(element).toEqualAttribute('aria-hidden', 'true');
-    expect(element).toHaveClass('is-hidden');
+    expect(element).not.toHaveClass('is-hidden');
```
dgonzalezr, line 39: I think this e2e test was good. But the new changes we were missing to set the `.is-hidden` class on componentDidLoad if `this.open = false`:

```ts
  componentDidLoad() {
    if (!this.open) {
      this.el.classList.add('is-hidden');
    }
  }
```

dgonzalezr (approved):

PR #819, opened 2024-02-06
feat(Notification): add `border` property to facilitate dynamic border-radius changes

Cata1989:
The purpose of this PR is to enhance by introducing dynamic border property to accommodate changes in border-radius.


packages/beeq/src/components/notification/bq-notification.tsx
```diff
@@ -231,8 +234,12 @@ export class BqNotification {
   // ===================================
 
   render() {
+    const style = {
+      ...(this.border && { '--bq-notification--border-radius': `var(--bq-radius--${this.border})` }),
+    };
```
dgonzalezr, line 239: Please leave an empty line between the const declaration and the return statement.

PR #835, opened 2024-02-14
refactor(Tokens)!: simplify the declarative design tokens

Cata1989:
BREAKING CHANGE: all the tokens related to `hover`, `active`, and `disable` states have been removed.

Before, we used to have e.g. `primary-hover`, and `primary-active` applied on a button when it hovered or active, this refactor relies now on the use of [color-mix()](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix) to archive the same result.


packages/beeq-tailwindcss/src/theme/default/light.ts
```diff
@@ -2,193 +2,194 @@ export const DefaultLightTheme = {
...
-  'bq-data--sky': 'var(--bq-sky-600)',
-  'bq-data--teal': 'var(--bq-teal-600)',
-  'bq-data--yellow': 'var(--bq-yellow-600)',
+  'bq-data--magenta': 'var(--bq-data-03)',
+  'bq-data--orange': 'var(--bq-data-06)',
+  'bq-data--purple': 'var(--bq-accent)',
+  'bq-data--red': 'var(--bq-danger)',
+  'bq-data--sky': 'var(--bq-sky-600)', // not found in root
+  'bq-data--teal': 'var(--bq-data-05)',
+  'bq-data--yellow': 'var(--bq-data-07)',
```
dgonzalezr, line 30: I'm confused here, if we are going to remove the data group from the light/dark files, why bother updating the value? 😉 The new ones will be those added in the theme root.ts file, right?

packages/beeq-tailwindcss/src/theme/default/dark.ts
```diff
@@ -17,8 +17,8 @@ export const DefaultDarkTheme = {
   'bq-icon--primary-hover': 'var(--bq-neutral-200)',
   'bq-icon--primary-active': 'var(--bq-neutral-50)',
   'bq-icon--primary-disabled': 'var(--bq-neutral-500)',
-  'bq-icon--primary-alt': 'var(--bq-white)',
-  'bq-icon--primary-alt-disabled': 'var(--bq-neutral-200)',
+  'bq-icon--alt': 'var(--bq-white)',
+  'bq-icon--alt-disabled': 'var(--bq-neutral-200)',
```
dgonzalezr, line 21: `bq-icon--alt-disable` should be removed, right?

dgonzalezr (approved):

PR #878, opened 2024-02-26
fix(Dropdown): disable the trigger element if the dropdown have `disable="true"`

Cata1989:
The purpose of this PR is to address the issue related to the disabled state management in the dropdown component. Previously, the disabled state of the dropdown was not properly synchronized with the associated button's disabled state. This PR fixes that issue by implementing a mechanism to correctly update the button's disabled state based on the dropdown's disabled property. As a result, the dropdown component now ensures consistent behavior when managing its disabled state, enhancing overall accessibility and user experience.

Before|After
-|-
Screenshot 2024-02-26 at 14 03 47|Screenshot 2024-02-26 at 14 04 29

### DIFF
--- a/packages/beeq/src/components/dropdown/bq-dropdown.tsx
+++ b/packages/beeq/src/components/dropdown/bq-dropdown.tsx
@@ -69,6 +69,14 @@ export class BqDropdown {
     this.bqOpen.emit({ open: this.open });
   }
 
+  @Watch('disabled')
+  handleDisabledChange(newValue: boolean) {
+    const bqButton = document.querySelector('bq-button');
+    if (!bqButton) return;
+
+    bqButton.disabled = newValue;
+  }
+
   // Events section
   // Requires JSDocs for public API documentation
   // ==============================================
@@ -82,6 +90,7 @@ export class BqDropdown {
 
   componentDidLoad() {
     this.triggerElem = this.el.querySelector('[slot="trigger"]');
+    this.handleDisabledChange(this.disabled);
   }
 
   // Listeners
--- a/packages/beeq/src/components/dropdown/bq-dropdown.tsx
+++ b/packages/beeq/src/components/dropdown/bq-dropdown.tsx
@@ -70,11 +70,11 @@ export class BqDropdown {
   }
 
   @Watch('disabled')
-  handleDisabledChange(newValue: boolean) {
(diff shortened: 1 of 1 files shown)


endv-bogdanb:
Not sure if this the right approach. I could use something else to trigger the dropdown. If we want this behaviour I would suggest to query by trigger slot and set disabled it's children.

dgonzalezr:
> Not sure if this the right approach. I could use something else to trigger the dropdown. If we want this behaviour I would suggest to query by trigger slot and set disabled it's children.

We have discussed about it, your concern is right, but in this case we would like only to cover the use case when the trigger is a `bq-button`. We could still set the disable attribute to whatever the trigger element is, I don't see any major issue if the attribute is set to an element that doesn't support disable as attribute.

packages/beeq/src/components/dropdown/bq-dropdown.tsx
```diff
@@ -69,6 +69,14 @@ export class BqDropdown {
...
   }
 
+  @Watch('disabled')
+  handleDisabledChange(newValue: boolean) {
+    const bqButton = document.querySelector('bq-button');
+    if (!bqButton) return;
+
+    bqButton.disabled = newValue;
+  }
+
```
dgonzalezr, line 79: Considering that we already have a `triggerElem` being set in the line:92,  maybe we could:

```ts
  @Watch('disabled')
  handleDisabledChange(newValue: boolean) {
    if (!this.triggerElem) return;
    // Optional: check if the triggerElem is a bq-button
    // if(!isHTMLElement(this.triggerElem, 'bq-button')) return;

    this.triggerElem?.setAttribute('disabled', 'true');
  }
```

cc. @endv-bogdanb

packages/beeq/src/components/dropdown/bq-dropdown.tsx
```diff
@@ -82,6 +90,7 @@ export class BqDropdown {
 
   componentDidLoad() {
     this.triggerElem = this.el.querySelector('[slot="trigger"]');
+    this.handleDisabledChange(this.disabled);
```
dgonzalezr, line 93:
```suggestion
    this.handleDisabledChange();
```

endv-bogdanb (approved):

PR #879, opened 2024-02-26
fix(Slider): add `not-allowed` cursor class for disabled state

Cata1989:
Screenshot 2024-02-26 at 15 47 20


endv-bogdanb (approved):

endv-bogdanb:
Should we consider thumbs ( dragging circles ) ? Currently they have cursors when disabled.

PR #946, opened 2024-03-19
feat(Progress): add new `bq-progress` component

Cata1989:
## Description

The purpose of this PR is to implement a `bq-progress` component.

## Related Issue

N/A

## Documentation

There's a Figma design here: https://www.figma.com/file/2VS47PTSkQWJkr7H6DpAdR/BEEQ-Design-System-(v1.0)?node-id=3991%3A109473&mode=dev

## Screenshots (if applicable)

There's a video with tooltip progress bar here: 
https://github.com/Endava/BEEQ/assets/98550729/0758ea64-72df-40e0-ad88-0fc312d7f931

## Checklist


packages/beeq/src/components/progress/bq-progress.tsx
```diff
@@ -0,0 +1,180 @@
...
+  @Prop({ reflect: true }) type: TProgressType = 'default';
+
+  /** If `true`, the progress bar will be displayed without border radius */
+  @Prop({ reflect: true }) level: boolean = false;
+
+  /** It `true`, the progress bar will be displayed with percentage text */
+  @Prop({ reflect: true }) percentage: boolean = false;
+
+  /** It `true`, the progress bar will be displayed with percentage tooltip */
+  @Prop({ reflect: true }) tooltip: boolean = false;
```
dgonzalezr, line 53: Do you think that `enableTooltip` could fit better in here?

packages/beeq/src/components/progress/bq-progress.tsx
```diff
@@ -0,0 +1,180 @@
...
+  // Public Methods must be async.
+  // Requires JSDocs for public API documentation.
+  // ===============================================
+
+  // Local methods
+  // Internal business logic.
+  // These methods cannot be called from the host element.
+  // =======================================================
+
+  private previousValue: number | null = null;
```
dgonzalezr, line 104: Shouldn't his be in the 'Own properties' section?

packages/beeq/src/components/progress/bq-progress.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+    return (
+      <div class="flex items-center gap-xs">
+        <div class="relative flex items-center">
+          <progress class={progressClasses} value={this.value} max="100" slot="triger"></progress>
+          {this.tooltip && (
+            <bq-tooltip
+              class="absolute"
+              exportparts="base,trigger,panel"
+              always-visible
```
dgonzalezr, line 161: Withing JSX you can bind to period instead: `alwaysVisible={true}`

packages/beeq/src/components/progress/bq-progress.tsx
```diff
@@ -0,0 +1,180 @@
...
+              style={{ left: `${this.value}%` }}
+            >
+              <div class="absolute h-1 w-1" slot="trigger"></div>
+              {this.value}
+            </bq-tooltip>
+          )}
+        </div>
+        {this.percentage && (
+          <div
+            class={{ 'font-medium leading-regular text-text-primary': true, 'text-ui-danger': this.type === 'error' }}
```
dgonzalezr, line 172: I think we need font-variant here

dgonzalezr (changes requested):

packages/beeq/src/components/progress/__tests__/bq-progress.e2e.ts
dgonzalezr, line 1: We need more e2e tests here, e.g:
- testing the label value appearance
- testing the tooltip rendering
- testing the error style

packages/beeq/src/components/progress/_storybook/bq-progress.stories.tsx
dgonzalezr, line 1: More storybook examples?
- Default
- Error
- With label
- With tooltip

packages/beeq/src/components/progress/bq-progress.tsx
```diff
@@ -0,0 +1,184 @@
...
+      'progress-bar__border-shape rounded-full': !this.borderShape,
+      'h-1': this.thickness === 'medium',
+      'h-2': this.thickness === 'large',
+      indeterminate: this.checkIsIndeterminated(),
+    };
+
+    return (
+      <div class="flex items-center gap-xs">
+        <div class="relative flex items-center">
+          <progress class={progressClasses} value={this.value} max="100" slot="triger"></progress>
```
dgonzalezr, line 156: Is there any reason why we should set a slot name here?

```suggestion
          <progress class={progressClasses} value={this.value} max="100"></progress>
```

Cata1989: Nope, I'll remove it. Added it just by mistake.

packages/beeq/src/components/progress/scss/bq-progress.variables.scss
```diff
@@ -0,0 +1,13 @@
...
+/*                   Progress custom properties                  */
+/* -------------------------------------------------------------------------- */
+
+:host {
+  /**
+   * @prop --bq-progress--width - Width of the progress
+   * @prop --bq-progress-left-offset-position - Default left offset tooltip position
+   */
+
+  --bq-progress--width: 250px;
```
dgonzalezr, line 11: We should let the component to adjust to his parent width rather than forcing the width.

packages/beeq/src/components/progress/bq-progress.tsx
```diff
@@ -0,0 +1,184 @@
...
+  @Prop({ reflect: true }) value = 0;
+
+  /** Progress bar thickness */
+  @Prop({ reflect: true }) thickness: TProgressThickness = 'medium';
+
+  /** Progress type */
+  @Prop({ reflect: true }) type: TProgressType = 'default';
+
+  /** If `true`, the progress bar will be displayed without border radius */
+  @Prop({ reflect: true }) borderShape: boolean = false;
```
dgonzalezr, line 49: Perhaps it will be easy to understand if the `borderShape` accept two values: `square` | `rounded`. With default value to be `rounded` as per design.

packages/beeq/src/components/progress/bq-progress.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+  // State() variables
+  // Inlined decorator, alphabetical order
+  // =======================================
+
+  // Public Property API
+  // ========================
+
+  /** It defines the mode of progress bar to display */
+  @Prop({ reflect: true }) mode: TProgressMode = 'determinate';
```
dgonzalezr, line 37: I would rather recommend to rename this property to `indeterminate` as a boolean (default `false`) and when `indeterminate` is set, we could just:

```jsx
<progress class={progressClasses} value={this.indeterminate ? undefined : this.value} max="100" />
...
{this.enableTooltip && !this.indeterminate && ( ... )}
...
<div
  style={{ fontVariant: 'tabular-nums' }}
  class={{
    ' font-medium tabular-nums leading-regular text-text-primary': true,
    'text-ui-danger': this.type === 'error',
    hidden: !this.label || this.indeterminate,
  }}
>
  {this.value}%
</div>
```

packages/beeq/src/components/progress/scss/bq-progress.variables.scss
dgonzalezr, line 1: We could also add custom properties like:
- `--bq-progress-bar--height`: the progress bar's height
- `--bq-progress-bar --trackColor`: the progress bar track area (the grey one)
- `--bq-progress-bar--indicatorColor`: the progress bar color (inside the track area)

packages/beeq/src/components/progress/scss/bq-progress.variables.scss
```diff
@@ -0,0 +1,13 @@
...
+  --bq-progress-left-offset-position: 0px;
```
dgonzalezr, line 12: I think it is safe to remove this custom property, right?

packages/beeq/src/components/progress/bq-progress.tsx
```diff
@@ -0,0 +1,158 @@
...
+  /** A number representing the current value of the progress bar */
+  @Prop({ reflect: true }) value = 0;
+
+  /** Progress bar thickness */
+  @Prop({ reflect: true }) thickness: TProgressThickness = 'medium';
+
+  /** Progress type */
+  @Prop({ reflect: true }) type: TProgressType = 'default';
+
+  /** If `rounded`, the progress bar will be displayed without border radius */
```
dgonzalezr, line 46:
```suggestion
  /** It will set the border style of the progress bar */
```

packages/beeq/src/components/progress/bq-progress.tsx
```diff
@@ -0,0 +1,158 @@
...
+  @Prop({ reflect: true }) borderShape: TProgressBorderShape = 'rounded';
+
+  /** If `true`, the progress bar will be displayed with percentage text */
```
dgonzalezr, line 49:
```suggestion
  /** If `true`, a label text showing the value (in percentage) will be shown */
```

packages/beeq/src/components/progress/bq-progress.tsx
```diff
@@ -0,0 +1,158 @@
...
+  @Prop({ reflect: true }) label: boolean = false;
+
+  /** If `true`, the progress bar will be displayed with percentage tooltip */
```
dgonzalezr, line 52:
```suggestion
  /** If `true`, a tooltip will be shown displaying the progress value */
```

packages/beeq/src/components/progress/bq-progress.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+  private validateValue(newValue: number) {
+    if (this.value) {
+      const clampedValue = Math.max(0, Math.min(100, newValue));
+      if (newValue !== clampedValue) {
+        this.value = clampedValue;
+      }
+    }
+    return this.value;
+  }
```
dgonzalezr, line 108: The method can be improved by removing the unnecessary `if` condition. The `if` condition checks if `this.value` is truthy, but it doesn't make sense in this context because we are validating the `newValue`, not `this.value`.

```suggestion
  private validateValue(newValue: number) {
    const clampedValue = Math.max(0, Math.min(100, newValue));
    if (newValue !== clampedValue) {
      this.value = clampedValue;
    }
    return this.value;
  }
```

packages/beeq/src/components/progress/bq-progress.tsx
```diff
@@ -0,0 +1,158 @@
...
+  // =======================================
+
+  // Public Property API
+  // ========================
+
+  /** If `true` the indeterminate state of progress bar is enabled */
+  @Prop({ reflect: true }) indeterminate: boolean = false;
+
+  /** A number representing the current value of the progress bar */
+  @Prop({ reflect: true }) value = 0;
```
dgonzalezr, line 38: You need to set this property as mutable since it can change its value internally in the `validateValue` method.

```suggestion
  @Prop({ reflect: true, mutable: true }) value = 0;
```

packages/beeq/src/components/progress/bq-progress.tsx
```diff
@@ -0,0 +1,158 @@
...
+    const style = {
+      ...(this.thickness === 'large' && { '--bq-progress-bar--height': '8px' }),
+      ...(this.type === 'error' && { '--bq-progress-bar--indicatorColor': 'var(--bq-ui--danger)' }),
+    };
+
+    return (
+      <Host style={style}>
+        <div class="flex items-center gap-xs">
+          <div class="relative flex w-full items-center">
+            <progress class={progressClasses} value={this.indeterminate ? undefined : this.value} max="100"></progress>
```
dgonzalezr, line 128: We can move the progress CSS classes to the element, right?

```suggestion
    const style = {
      ...(this.thickness === 'large' && { '--bq-progress-bar--height': '8px' }),
      ...(this.type === 'error' && { '--bq-progress-bar--indicatorColor': 'var(--bq-ui--danger)' }),
    };

    return (
      <Host style={style}>
        <div class="flex items-center gap-xs">
          <div class="relative flex w-full items-center">
            <progress
              class={{
                [`progress-bar progress-bar__${this.type} ${this.thickness}`]: true,
                'progress-bar__border-shape rounded-full': this.borderShape === 'rounded',
              }}
              value={this.indeterminate ? undefined : this.value}
              max="100"
            ></progress>
```

packages/beeq/src/components/progress/bq-progress.tsx
```diff
@@ -0,0 +1,158 @@
...
+              style={{ fontVariant: 'tabular-nums' }}
+              class={{
+                'font-medium leading-regular text-text-primary': true,
+                'text-ui-danger': this.type === 'error',
+                hidden: !this.label || this.indeterminate,
+              }}
+            >
+              {this.value}%
+            </div>
+          )}
```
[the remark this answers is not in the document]
Cata1989: Yes, we can.

packages/beeq/src/components/progress/bq-progress.tsx
Cata1989: Yes, we can .

packages/beeq/src/components/progress/bq-progress.tsx
dgonzalezr, line 1: We are still missing exposed shadow DOM parts, please make sure to add them and document them on top of the @Component decorator 🙂

packages/beeq/src/components/progress/_storybook/bq-progress.stories.tsx
```diff
@@ -0,0 +1,91 @@
...
+`;
+
+export const Default: Story = {
+  render: Template,
+  args: {
+    value: 10,
+  },
+};
+
+export const MediumThickness: Story = {
```
dgonzalezr, line 59: The name of this story is misleading, we're showing here the large-thickness example

```suggestion
export const LargeThickness: Story = {
```

packages/beeq/src/components/progress/_storybook/bq-progress.stories.tsx
```diff
@@ -0,0 +1,91 @@
...
+    label: true,
+  },
+};
+
+export const WithTooltip: Story = {
+  render: Template,
+  args: {
+    value: 80,
+    'enable-tooltip': true,
+    'border-shape': 'square',
```
dgonzalezr, line 89: I would recommend a separate Story example for the border shape instead of mixing it with the tooltip. Maybe You could add the example above this one and leave this as a combined example then 😉 what do you think?

Cata1989: Yeah, sounds good!

packages/beeq/src/components/progress/__tests__/bq-progress.e2e.ts
```diff
@@ -0,0 +1,58 @@
...
+    });
+    const element = await page.find('bq-progress');
+
+    expect(element.shadowRoot).not.toBeNull();
+  });
+
+  it('should render the progress bar with label', async () => {
+    const page = await newE2EPage({
+      html: '<bq-progress label></progress>',
+    });
```
dgonzalezr, line 26: Perhaps we need to set the value so that we can test that the label has a text equal to `${value}%`, would that make sense?

Cata1989: Yes, good idea.

packages/beeq/src/components/progress/__tests__/bq-progress.e2e.ts
dgonzalezr, line 1: I would recommend adding value to most of the e2e tests and (if possible) having a test that checks when there's no value set, the progress indicator has no width, or is not shown.

packages/beeq/src/components/progress/__tests__/bq-progress.e2e.ts
```diff
@@ -0,0 +1,58 @@
+import { newE2EPage } from '@stencil/core/testing';
+// import { computedStyle } from '../../../shared/test-utils';
```
dgonzalezr, line 2: Please, remove this if it's not needed

packages/beeq/src/components/progress/bq-progress.tsx
```diff
@@ -0,0 +1,157 @@
...
+        <div class="flex items-center gap-xs">
+          <div class="relative flex w-full items-center">
+            <progress
+              class={{
+                [`progress-bar progress-bar__${this.type} ${this.thickness}`]: true,
+                'progress-bar__border-shape rounded-full': this.borderShape === 'rounded',
+              }}
+              value={this.indeterminate ? undefined : this.value}
+              max="100"
+            ></progress>
```
dgonzalezr, line 129: You can use self close tag with JSX when there is no child

```suggestion
            <progress
              class={{
                [`progress-bar progress-bar__${this.type} ${this.thickness}`]: true,
                'progress-bar__border-shape rounded-full': this.borderShape === 'rounded',
              }}
              value={this.indeterminate ? undefined : this.value}
              max="100"
            />
```

packages/beeq/src/components/progress/__tests__/bq-progress.e2e.ts
```diff
@@ -0,0 +1,90 @@
...
+      ),
+    );
+    expect(indicatorColor).toEqual(uiErrorColor);
+  });
+
+  it('should render the progress bar in indeterminate mode', async () => {
+    const value = 80;
+    const page = await newE2EPage({
+      html: `<bq-progress value="${value}" indeterminate></progress>`,
+    });
```
dgonzalezr, line 85: Since we're doing nothing with the value, we can just set it directly.

```suggestion
    const page = await newE2EPage({
      html: `<bq-progress value="80" indeterminate></progress>`,
    });
```

packages/beeq/src/components/progress/__tests__/bq-progress.e2e.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    const element = await page.find('bq-progress >>> [part="label"] span');
+    expect(element).not.toBeNull();
+    expect(element.textContent).toEqual(`${value}%`);
+  });
+
+  it('should render the progress bar without label', async () => {
+    const page = await newE2EPage({
+      html: '<bq-progress label="false"></progress>',
```
dgonzalezr, line 35:
```suggestion
      html: '<bq-progress value="60" label="false"></progress>',
```

dgonzalezr (approved):

PR #1029, opened 2024-04-22
feat(Date Picker): add new `bq-date-picker` component

### LINKED ISSUE #916: Feat: Date picker
### Additional Information

Figma: https://www.figma.com/file/2VS47PTSkQWJkr7H6DpAdR/BEEQ-Design-System-(v1.0)?type=design&node-id=3991%3A54950&mode=design&t=IM6eKno8O3mv7ODZ-1

Cata1989:
## Description

The purpose of this PR is to introduce a new feature like Date Picker which uses an external library named Cally (https://wicky.nillia.ms/cally/) which is not integrated in our Beeq Components. 

## Related Issue

Fixes #916 

## Checklist


packages/beeq/src/components/date-picker/scss/bq-date-picker.variables.scss
```diff
@@ -0,0 +1,64 @@
...
+
+  --bq-date-picker--icon-size: 24px;
+
+  --bq-date-picker--label-margin-bottom: theme('spacing.xs');
+  --bq-date-picker--label-text-size: theme('fontSize.s');
+  --bq-date-picker--label-text-color: theme('colors.text.primary');
+  --bq-date-picker--currentDate-text-color: theme('colors.text.brand');
+
+  --bq-date-picker--pading-start: theme('spacing.m');
+  --bq-date-picker--pading-end: theme('spacing.m');
```
dgonzalezr, line 56: There is a typo here, I think it should be:

```suggestion
  --bq-date-picker--padding-start: theme('spacing.m');
  --bq-date-picker--padding-end: theme('spacing.m');
```

Be aware of updating its usage on `bq-date-picker.scss` and the @prop documentation at the top of this file

packages/beeq/src/components/date-picker/scss/bq-date-picker.variables.scss
```diff
@@ -0,0 +1,64 @@
...
+  --bq-date-picker--border-radius: theme('borderRadius.s');
+  --bq-date-picker--border-width: 1px;
+  --bq-date-picker--currentDate-border-width: 2px;
+  --bq-date-picker--border-style: solid;
+
+  --bq-date-picker--gap: theme('spacing.xs');
+
+  --bq-date-picker--helper-margin-top: theme('spacing.xs');
+  --bq-date-picker--helper-text-size: theme('fontSize.s');
+  --bq-date-picker--helper-text-color: theme('colors.text.primary');
```
dgonzalezr, line 46: I don't see any usage of these variables, we should probably remove them.

packages/beeq/src/components/panel/bq-panel.tsx
```diff
@@ -138,10 +138,12 @@ export class BqPanel {
   render() {
     return (
       <div
-        class="bq-panel"
+        class={{
+          'bq-panel': true,
+          'panel-hidden': !this.open,
+        }}
```
dgonzalezr, line 144: This affects everything that uses the bq-panel, e.g. the dropdown, and the select, where the panel in hidden mode (not open) seems to visually fade out when the component renders, normally the panel shouldn't be visible.

https://github.com/Endava/BEEQ/assets/328492/cdce9321-1f5d-48c4-9b64-33a50c86e303

packages/beeq/src/components/date-picker/bq-date-picker.tsx
```diff
@@ -0,0 +1,500 @@
...
+    const labelId = `bq-date-picker__label-${this.name || this.fallbackInputId}`;
+
+    const componentTypes = {
+      multi: 'calendar-multi',
+      range: 'calendar-range',
+      default: 'calendar-date',
+    };
+
+    const key = (this.multi && 'multi') || (this.range && 'range') || 'default';
+    const CalendarComponentType = componentTypes[key];
```
dgonzalezr, line 374: This is very nice, well thought 👏🏼🙂 

We can also improve it a bit more if we move the lookup object to a getter and return the corresponding type from there. Something like:

```js
private get CalendarType() {
  // Define a lookup object to map properties to component types
  const componentTypes = {
    multi: 'calendar-multi',
    range: 'calendar-range',
    default: 'calendar-date',
  };
  // Find the first property (multi or range) that is truthy
  const type = ['multi', 'range'].find((t) => this[t]);
  // Return the corresponding component type, or the default type if no truthy property was found
  return componentTypes[type] || componentTypes.default;
}

render() {
  ...
  const CallyCalendar = this.CalendarType;

  ...
  <CallyCalendar
    ...
  >
    ...
  </CallyCalendar>
}
```

What do you think?

packages/beeq/src/components/date-picker/bq-date-picker.tsx
```diff
@@ -0,0 +1,500 @@
...
+              min={this.min}
+              max={this.max}
+              months={this.months}
+              focusedDate={this.processFocusedDateValue(this.value)}
+              firstDayOfWeek={this.firstDayOfWeek}
+              showOutsideDays={this.showOutsideDays}
+              onChange={(ev: { target: { value: string } }) => {
+                this.value = ev.target.value;
+                this.open = !!this.multi;
+              }}
```
dgonzalezr, line 487: I would recommend to move this into a private method

```ts
  private handleCalendarChange = (ev: { target: { value: string } }) => {
    this.value = ev.target.value;
    this.open = !!this.multi;
  };
  ...
  onChange={this.handleCalendarChange}
```

dgonzalezr: Also, when the user selects a date from the calendar, the bqChange event is not emitted, meaning that we should also dispatch that event in the onChange of the Calendar.

packages/beeq/src/global/scripts/global.ts
```diff
@@ -0,0 +1 @@
+import 'cally';
```
endv-bogdanb, line 1: I think we should add their components to `JSXBase` so we get typescript errors.
```suggestion
import 'cally';
import type { JSXBase } from '@stencil/core/internal';
import type { CalendarDateProps, CalendarMonthProps, CalendarMultiProps, CalendarRangeProps } from 'cally';

declare module '@stencil/core' {
  // eslint-disable-next-line @typescript-eslint/no-namespace
  export namespace JSX {
    interface IntrinsicElements {
      'calendar-multi': CalendarMultiProps & JSXBase.HTMLAttributes<CalendarMultiProps>;
      'calendar-range': CalendarRangeProps & JSXBase.HTMLAttributes<CalendarDateProps>;
      'calendar-date': CalendarDateProps & JSXBase.HTMLAttributes<CalendarDateProps>;
      'calendar-month': CalendarMonthProps & JSXBase.HTMLAttributes<CalendarMonthProps>
    }
  }
}

```

packages/beeq/src/components/date-picker/bq-date-picker.tsx
endv-bogdanb: I would add `as const` for `componentTypes` and type `array` so the return is properly typed.

packages/beeq/src/components/date-picker/bq-date-picker.tsx
```diff
@@ -0,0 +1,500 @@
...
+    const buttonExportParts = 'button,day,selected,today,disallowed,outside,range-start,range-end,range-inner';
+
+    if (this.range || (this.multi && this.months)) {
+      for (let i = 0; i < this.months; i++) {
+        const offset = i > 0 ? i : undefined;
+        const className = offset ? 'hidden sm:block' : '';
+        months.push(
+          <calendar-month
+            offset={offset}
+            className={className}
```
endv-bogdanb, line 327: Should this be `class` ?

packages/beeq/src/components/date-picker/bq-date-picker.tsx
```diff
@@ -0,0 +1,500 @@
...
+  @Prop({ reflect: true }) skidding?: number = 0;
+
+  /** Defines the strategy to position the Date picker panel */
+  @Prop({ reflect: true }) strategy?: 'fixed' | 'absolute' = 'fixed';
+
+  /** If `true`, the Date picker panel will accepts more than 1 month to display */
+  @Prop({ reflect: true }) range: boolean = false;
+
+  /** If `true`, the Date picker panel will accepts to select multiple individual dates */
+  @Prop({ reflect: true }) multi: boolean = false;
```
dgonzalezr, line 129: What do you think about using a single property (e.g. `type`) to set which variant of the calendar to use?
Because we cannot mix `range` with `multi` for example, so probably will be better if we just have?

```js
/** It defines how the calendar will behave, allowing single date selection, range selection, or multiple date selection */
@Prop({ reflect: true }) type: TDatePickerType = 'single';
```
where `TDatePickerType` is
```js
export const DATE_PICKER_TYPE = ['single', 'mullti', 'range'] as const;
export type TDatePickerType = (typeof DATE_PICKER_TYPE)[number];
```

packages/beeq/src/components/date-picker/_storybook/bq-date-picker.stories.tsx
```diff
@@ -0,0 +1,193 @@
...
+    skidding: 0,
+    strategy: 'absolute',
+    required: false,
+    'validation-status': 'none',
+    range: false,
+    multi: false,
+    months: 1,
+    value: undefined,
+    isDateDisallowed: undefined,
+    customDisallowedDate: undefined,
```
dgonzalezr, line 95: I cannot see this prop being defined and used in the implementation of the component (_in bq-date-picker.tsx_) 🤔

packages/beeq/src/components/date-picker/_storybook/bq-date-picker.stories.tsx
```diff
@@ -0,0 +1,193 @@
...
+    bqBlur: { action: 'bqBlur' },
+    bqChange: { action: 'bqChange' },
+    bqClear: { action: 'bqClear' },
+    bqFocus: { action: 'bqFocus' },
+    bqInput: { action: 'bqInput' },
+    // Not part of the public API, so we don't want to expose it in the docs
+    noLabel: { control: 'boolean', table: { disable: true } },
+    prefix: { control: 'boolean', table: { disable: true } },
+    suffix: { control: 'boolean', table: { disable: true } },
+    customDisallowedDate: { control: 'text' },
```
dgonzalezr, line 67: I cannot see this prop being defined and used in the implementation of the component (_in bq-date-picker.tsx_) 🤔

packages/beeq/src/components/date-picker/bq-date-picker.tsx
```diff
@@ -0,0 +1,500 @@
...
+   * - `'none'`: No validation status is set.
+   * - `'error'`: The input has a validation error.
+   * - `'warning'`: The input has a validation warning.
+   * - `'success'`: The input has passed validation.
+   */
+  @Prop({ reflect: true }) validationStatus: TInputValidation = 'none';
+
+  /** The select input value represents the currently selected date or range and can be used to reset the field to a previous value.
+   * All dates are expected in ISO-8601 format (YYYY-MM-DD). */
+  @Prop({ reflect: true, mutable: true }) value: TInputValue;
```
dgonzalezr, line 151: The value prop should be string type only, keep in mind that we're dealing with dates here 😉 

```suggestion
  @Prop({ reflect: true, mutable: true }) value: string;
```

packages/beeq/src/components/date-picker/bq-date-picker.tsx
```diff
@@ -0,0 +1,500 @@
...
+ * @part tr - Any row within the table.
+ * @part head - The table's header row.
+ * @part week - The table's body rows.
+ * @part th - The table's header cells.
+ * @part td - The table's body cells.
+ * @part button - Any button used in the component.
+ * @part day - The buttons corresponding to each day in the grid.
+ * @part selected - Any days which are selected. 
+ * @part today - Today's day.
+ * @part disallowed - Any daythat has been disallowed via isDateDisallowed.
```
dgonzalezr, line 39:
```suggestion
 * @part disallowed - Any day that has been disallowed via isDateDisallowed.
```

packages/beeq/src/components/date-picker/bq-date-picker.tsx
```diff
@@ -0,0 +1,545 @@
...
+    if (!value) return;
+
+    const dateLength = 10; // Length of a standard date in the format YYYY-MM-DD
+    return value.slice(-dateLength);
+  };
+
+  private formatDate = (value: string): string | undefined => {
+    if (!value) return;
+
+    const dateFormatter = new Intl.DateTimeFormat(this.locale, this.formatOptions);
```
endv-bogdanb, line 389: Not sure if this would work, since `formatOptions` is object we need to `JSON.stringify`.

packages/beeq/src/components/date-picker/bq-date-picker.tsx
```diff
@@ -0,0 +1,545 @@
...
+  };
+
+  private handleChange = (ev: Event) => {
+    if (this.disabled) return;
+
+    if (!isHTMLElement(ev.target, 'input')) return;
+
+    const dateValue = new Date(ev.target.value);
+    if (!isNaN(dateValue.getTime())) {
+      this.value = new Intl.DateTimeFormat('fr-CA', { year: 'numeric', month: '2-digit', day: '2-digit' }).format(
```
endv-bogdanb, line 291: Why are we using `french canada` locale here ?

packages/beeq/src/components/date-picker/bq-date-picker.tsx
```diff
@@ -0,0 +1,545 @@
...
+    this.value = ev.target.value;
+  };
+
+  private handleChange = (ev: Event) => {
+    if (this.disabled) return;
+
+    if (!isHTMLElement(ev.target, 'input')) return;
+
+    const dateValue = new Date(ev.target.value);
+    if (!isNaN(dateValue.getTime())) {
```
endv-bogdanb, line 290: Don't think `getTime()` can be `NaN` unless date is invalid. I would go for `!isNaN(+dateValue)`.

packages/beeq/src/components/date-picker/bq-date-picker.tsx
Cata1989: Yes, we need to keep activ this format yyyy-mm-dd as well.

packages/beeq/src/components/date-picker/bq-date-picker.tsx
dgonzalezr: I think that's the idea, to trigger only the context inside the conditional if the string value is a valid date.

packages/beeq/src/components/date-picker/bq-date-picker.tsx
dgonzalezr: I think this is fine, it works with Object:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat#using_options
and the `formatOptions` property has been defined as Object only, meaning that it cannot be set via HTML attributes.

packages/beeq/src/components/date-picker/bq-date-picker.tsx
endv-bogdanb: I would use here [formatToParts](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/formatToParts#description) with a `filter` and `join` to construct format `yyyy-mm-dd`

dgonzalezr: Indeed, but you need something like  the following:

```js
const dateTimeFormat = new Intl.DateTimeFormat(this.locale, {
  year: 'numeric',
  month: '2-digit',
  day: '2-digit',
});
const dateParts = dateTimeFormat.formatToParts(dateValue);
const dateObject: { year?: string; month?: string; day?: string } = dateParts.reduce((acc, part) => {
  if (part.type === 'year' || part.type === 'month' || part.type === 'day') {
    acc[part.type] = part.value;
  }
  return acc;
}, {});
this.value = `${dateObject.year}-${dateObject.month}-${dateObject.day}`;
```
wondering if it is worth tho 🤔 

We could also simplify this by just doing:

```js
if (!isNaN(dateValue.getTime())) {
  this.value = dateValue.toLocaleDateString('fr-CA');
  this.formattedDate = this.formatDate(this.value);
  this.bqChange.emit({ value: this.value, el: this.el });
}
```

packages/beeq/src/components/date-picker/_storybook/bq-date-picker.stories.tsx
```diff
@@ -0,0 +1,296 @@
...
+      `;
+  const style = args.hasLabelTooltip
+    ? html`
+        <style>
+          bq-input {
+            width: 75vw;
+          }
+        </style>
+      `
+    : nothing;
```
dgonzalezr, line 137: I think this has no effect since you're hardcoding the parent div container that holds the `bq-date-picker` component.

```html
    <div class="w-[280px]">
      <bq-date-picker
```

packages/beeq/src/components/date-picker/_storybook/bq-date-picker.stories.tsx
```diff
@@ -0,0 +1,296 @@
...
+  },
+};
+
+export const Tooltip: Story = {
+  name: 'Label with "Info tooltip"',
+  render: Template,
+  args: {
+    hasLabelTooltip: true,
+  },
+};
```
dgonzalezr, line 288: For this specific example, I would recommend to add also:

```js
  parameters: {
    layout: 'centered',
  },
```

to the Tooltip story object.

packages/beeq/src/components/date-picker/bq-date-picker.tsx
```diff
@@ -0,0 +1,544 @@
...
+
+    this.bqFocus.emit(this.el);
+  };
+
+  private handleInput = (ev: Event) => {
+    if (this.disabled) return;
+
+    if (!isHTMLElement(ev.target, 'input')) return;
+    this.value = ev.target.value;
+  };
```
dgonzalezr, line 282: Since we drop `bqInput` event emitter, we can also remove this method and its usage.

packages/beeq/src/components/date-picker/_storybook/bq-date-picker.stories.tsx
```diff
@@ -0,0 +1,288 @@
...
+    if (!args.customDisallowedDate) {
+      return false;
+    }
+
+    const dateString = dateToIsoString(date);
+    return args.customDisallowedDate.split(',').some((date: string) => date.trim() === dateString);
+  };
+
+  return html`
+    <div class="w-[280px]">
```
dgonzalezr, line 148: I would rather remove this and let the component take the whole width space available.

packages/beeq/src/components/date-picker/_storybook/bq-date-picker.stories.tsx
```diff
@@ -0,0 +1,288 @@
...
+  render: Template,
+  args: {
+    disabled: true,
+  },
+};
+
+export const ValidationStatus: Story = {
+  name: 'Validation',
+  render: (args) => html`
+    <div class="flex flex-row gap-4">
```
dgonzalezr, line 264: We can use a grid instead

```suggestion
    <div class="grid grid-cols-1 gap-m sm:grid-cols-3">
```

packages/beeq/src/components/date-picker/_storybook/bq-date-picker.stories.tsx
```diff
@@ -0,0 +1,309 @@
...
+  render: Template,
+  args: {
+    type: 'multi',
+    months: 2,
+  },
+};
+
+export const InitialValue: Story = {
+  render: (args) => html`
+    <div class="flex flex-row gap-4">
```
dgonzalezr, line 210: You can use a grid instead:
```suggestion
    <div class="grid grid-cols-1 gap-m sm:grid-cols-3">
```
this will guarantee a responsive layout.

packages/beeq/src/components/date-picker/_storybook/bq-date-picker.stories.tsx
dgonzalezr: @Cata1989 I think this 👆🏼  is not solved yet

packages/beeq/src/components/date-picker/scss/bq-date-picker.scss
```diff
@@ -0,0 +1,240 @@
...
+  &::part(button day selected) {
+    @apply bg-[--bq-date-picker--range-background-color] hover:bg-hover-ui-brand;
+  }
+
+  &::part(outside) {
+    @apply text-[--bq-text--secondary] opacity-100;
+  }
+
+  &::part(selected) {
+    @apply rounded-[--bq-radius--s] text-[color:var(--bq-text--inverse)];
```
dgonzalezr, line 198: Use `--bq-text--alt` for accessibility contrast

```suggestion
    @apply rounded-[--bq-radius--s] text-[color:var(--bq-text--alt)];
```

packages/beeq/src/components/date-picker/bq-date-picker.tsx
```diff
@@ -0,0 +1,537 @@
...
+   *
+   * @param value - The value to be processed, can be a string.
+   * @returns The extracted last date portion of the value.
+   */
+  private focusedDate = (value: string) => {
+    if (!value) return;
+
+    const dateLength = 10; // Length of a standard date in the format YYYY-MM-DD
+    return value.slice(-dateLength);
+  };
```
dgonzalezr, line 377: We need to set the `focusedDate` to the first instance of `yyyy-mm-dd` of the value string, otherwise in a range or multi variant the user can not see where the initial value of the range is set:

CleanShot 2024-05-20 at 18 47 39@2x

```suggestion
  /**
   * Extracts and returns the first date part from a given string.
   * When the type of the date picker is 'range' or 'multi', the first or initial date part of the value
   * should be the focused date in the calendar.
   *
   * @param value - The value to be processed, can be a string.
   * @returns The extracted last date portion of the value.
   */
  private focusedDate = (value: string) => {
    if (!value) return;

    const dateRegex = /\b\d{4}-\d{2}-\d{2}\b/;
    const match = dateRegex.exec(value);
    return match ? match[0] : null;
  };
```

CleanShot 2024-05-20 at 18 51 28@2x

packages/beeq/src/components/date-picker/bq-date-picker.tsx
```diff
@@ -0,0 +1,537 @@
...
+            </span>
+          </div>
+          <div class="flex items-center justify-center">
+            <CallyCalendar
+              isDateDisallowed={this.isDateDisallowed}
+              locale={this.locale as string}
+              value={this.value}
+              min={this.min}
+              max={this.max}
+              months={this.months}
```
dgonzalezr, line 520: I would recommend tying this to a different prop (e.g.: `monthsPerView`)
```js
/** The number of months to display per page when using next/previous buttons. /*
@Prop({ reflect: true }) monthsPerView: number = 1;
```

packages/beeq/src/components/date-picker/scss/bq-date-picker.variables.scss
```diff
@@ -48,6 +48,7 @@
   --bq-date-picker--padding-end: theme('spacing.m');
   --bq-date-picker--paddingY: theme('spacing.s');
 
+  --bq-date-picker--day-size: theme(spacing.xl);
```
Cata1989, line 51: I think that here should be `--bq-date-picker--day-size: theme('spacing.xl');` isn't it?

dgonzalezr (dismissed):

packages/beeq/src/components/date-picker/scss/bq-date-picker.variables.scss
Cata1989: Ok, got it!

dgonzalezr (approved):

PR #1074, opened 2024-05-29
fix(Date Picker): clear incomplete range selection

Cata1989:
## Description

The purpose of this PR is to update the Cally library and fix some bugs. - (reset user's start date selection on calendar-range and ability to page by one month at a time or by months - pagination)

## Related Issue

https://github.com/WickyNilliams/cally/issues/48
https://github.com/WickyNilliams/cally/issues/46

Fixes #ISSUE_NUMBER

## Screenshots (if applicable)

https://github.com/Endava/BEEQ/assets/98550729/2eff1fc7-ef14-42b3-bc51-70384c81c4da

https://github.com/Endava/BEEQ/assets/98550729/c70eed18-5fad-4003-8738-661541a44a26

## Checklist


packages/beeq/src/components/date-picker/bq-date-picker.tsx
```diff
@@ -80,6 +81,8 @@ export class BqDatePicker {
   @State() hasPrefix = false;
   @State() hasSuffix = false;
   @State() hasValue = false;
+  @State() hasRangeEnd = false;
+  @State() setTentative = null;
```
dgonzalezr, line 85: We should support passing an external value via prop/attribute since Cally supports that, meaning that instead  of adding a state, we could provide a mutable property:

```ts
/** The date that is tentatively selected e.g. the start of a range selection  */
  @Prop({ reflect: true, mutable: true }) tentative?: string;
```

packages/beeq/src/components/date-picker/bq-date-picker.tsx
```diff
@@ -239,6 +246,22 @@ export class BqDatePicker {
...
+
+    const { top, bottom, left, right } = datePickerElement.getBoundingClientRect();
+    const { clientY, clientX } = event;
+
+    const isOutside = clientY < top || clientY > bottom || clientX < left || clientX > right;
+
+    if (isOutside && !hasRangeEnd) {
+      this.setTentative = this.setTentative === undefined ? null : undefined;
+    }
+  }
```
dgonzalezr, line 263: We could simplify this by using the `isEventTargetChildOfElement` utility from `packages/beeq/src/shared/utils/event.ts`, and I will recommend listening for the `click` event instead of `mousedown`:

```suggestion
  @Listen('click', { target: 'window', capture: true })
  async handleMouseClick(event: MouseEvent) {
    const { open, datePickerElement, type, hasRangeEnd } = this;
    if (!open || !datePickerElement || type !== 'range' || event.button !== 0) return;

    if (!isEventTargetChildOfElement(event, this.el) && !hasRangeEnd) {
      this.tentative = undefined;
    }
  }
```

packages/beeq/src/components/date-picker/bq-date-picker.tsx
```diff
@@ -305,11 +328,20 @@ export class BqDatePicker {
     this.open = this.type === 'multi';
   };
 
+  private handleCalendarRangeStart = () => {
+    this.hasRangeEnd = false;
+  };
```
dgonzalezr, line 333: If we add a new `tentative` property, we should guarantee that `tentative` gets the range start value

```suggestion
  private handleCalendarRangeStart = (ev: CustomEvent) => {
    this.hasRangeEnd = false;
    this.tentative = new Date(ev.detail).toLocaleDateString('fr-CA');
  };
```

packages/beeq/src/components/date-picker/bq-date-picker.tsx
```diff
@@ -515,18 +547,25 @@ export class BqDatePicker {
...
+          >
             <CallyCalendar
               isDateDisallowed={this.isDateDisallowed}
               locale={this.locale as string}
               value={this.value}
               min={this.min}
               max={this.max}
-              months={this.monthsPerView}
+              months={this.months}
+              tentative={this.setTentative}
```
dgonzalezr, line 561:
```suggestion
              tentative={this.tentative}
```

packages/beeq/src/components/date-picker/bq-date-picker.tsx
dgonzalezr: Don't forget to update Storybook with the new attribute.

packages/beeq/src/components/date-picker/bq-date-picker.tsx
dgonzalezr: By assigning `undefined`, you can see that it is enough to clear the `tentative` value.

dgonzalezr: Keep consistency when using the `event` argument. In this file, all the methods use `ev` instead of `event`.

dgonzalezr (approved):

PR #1085, opened 2024-06-03
feat(Page Title): add new `bq-page-title` component

Cata1989:
## Description

The purpose of this PR is to add a Pate Title component, based on our Figma design here: https://www.figma.com/design/2VS47PTSkQWJkr7H6DpAdR/BEEQ-Design-System-(v1.0)?node-id=3991-104418&m=dev

## Related Issue

Fixes #ISSUE_NUMBER

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/Endava/BEEQ/blob/main/CONTRIBUTING.md) document.
- [ ] I have read the [CODE OF CONDUCT](https://github.com/Endava/BEEQ/blob/main/CODE_OF_CONDUCT.md) document.
- [ ] I have reviewed my own code.
- [ ] I have tested the changes locally.
- [ ] I have added unit tests and e2e tests (if applicable).
- [ ] I have requested reviews from relevant team members.


packages/beeq/src/components/page-title/bq-page-title.tsx
```diff
@@ -0,0 +1,126 @@
...
+  // =======================================
+
+  // Public Property API
+  // ========================
+
+  // Prop lifecycle events
+  // =======================
+
+  /** If true, the page title back icon will be shown */
+  @Prop({ reflect: true }) showBackIcon: boolean;
```
dgonzalezr, line 34: Can we rename this propery to something like: `haveBackNavigation` or something similar?
What do you think?

packages/beeq/src/components/page-title/bq-page-title.tsx
```diff
@@ -0,0 +1,126 @@
...
+                <bq-icon
+                  color="text--brand"
+                  name="download-simple"
+                  size="24"
+                  weight="bold"
+                  part="icon"
+                  exportparts="base,svg"
+                />
+              </slot>
+            </div>
```
dgonzalezr, line 110: After a discussion with UX, it seems that actions can be at the end but also in front of the title content, considering that scenario it would make sense to handle prefix and suffix slots for actions instead of just `actions` at the end 😉 . In other words, from project side we could have somethig like:

```html
<bq-page-title>
  <div slot="prefix">
    <!-- button icon actions here -->
  </div>
  Title
</bq-page-title>
```
or
```html
<bq-page-title>
  Title
  <div slot="suffix">
    <!-- button icon actions here -->
  </div>
</bq-page-title>
```

packages/beeq/src/components/page-title/bq-page-title.tsx
```diff
@@ -0,0 +1,126 @@
...
+          </div>
+          {/* SUB-TITLE */}
+          <div
+            class={{
+              'title-font text-l font-medium leading-regular text-text-secondary': true,
+              hidden: !this.showSubTitle,
```
dgonzalezr, line 116: We should hide the container if there's no slot content for `sub-title`, I think there's no need for a property. What your thoughts on this?

packages/beeq/src/components/page-title/bq-page-title.tsx
```diff
@@ -0,0 +1,126 @@
...
+  // =======================
+
+  /** If true, the page title back icon will be shown */
+  @Prop({ reflect: true }) showBackIcon: boolean;
+
+  /** If true, the page title back icon will be shown */
+  @Prop({ reflect: true }) showActionIcons: boolean;
+
+  /** If true, the sub title of page title will be shown */
+  @Prop({ reflect: true }) showSubTitle: boolean;
```
dgonzalezr, line 40: Not sure if we need these properties.  For the actions, we don't need to support any specific action icon, we don't which actions projects will have and we cannot assume those rendered in the `actions` slot as the default. That said, I think that handling the slots and displaying them when they have content would be enough.

packages/beeq/src/components/page-title/scss/bq-page-title.variables.scss
```diff
@@ -0,0 +1,10 @@
+/* -------------------------------------------------------------------------- */
+/*                   Page title custom properties                  */
+/* -------------------------------------------------------------------------- */
+
+:host {
+  /**
+   * @prop --bq-page-title--margin - Page title margin
+   */
+  --bq-page-title--margin: theme(spacing.m);
```
dgonzalezr, line 9: Instead of adding a CSS custom property for margin, I would recommend adding:
```css
--text-color
--text-lineHeight
--text-size
--text-weight
```
which are more relevant to the scope of the component.

packages/beeq/src/components/page-title/bq-page-title.tsx
```diff
@@ -0,0 +1,126 @@
...
+        <div class={{ 'flex p-s': true, hidden: !this.showBackIcon }} part="back">
+          <slot name="back">
+            <bq-icon
+              color="text--primary"
+              name="arrow-left"
+              size="24"
+              weight="bold"
+              part="icon"
+              exportparts="base,svg"
+            />
```
dgonzalezr, line 82: The back navigation should handle accessibility, meaning this should be a button icon instead of just an icon.

dgonzalezr: Also, in case consumers use the default back button navigation, we should provide a callback handler to allow consumers to trigger any type of action they would like to execute when this button is clicked, right?

endv-bogdanb: I think we should handle this as `prefix` / `suffix` like button.  We have [default title](https://www.figma.com/design/2VS47PTSkQWJkr7H6DpAdR/BEEQ-Design-System-(v1.0)?node-id=3991-104418&t=jiu69tRNYcC905Zl-0) without any adornments.

dgonzalezr: Yes, you are right, but besides handling the prefix/suffix as you suggested, I'm also thinking of providing an easy way to use back navigation and save some time for developers to set up the layout. What do you guys think, @endv-bogdanb, @Cata1989

packages/beeq/src/components/page-title/bq-page-title.tsx
```diff
@@ -0,0 +1,170 @@
...
+        {/* Back navigation button */}
+        <div class={{ flex: true, '!hidden': !this.haveBackNavigation }} part="back">
+          <slot name="back">
+            <bq-button
+              appearance="link"
+              part="btn-back"
+              exportparts="button"
+              onBlur={this.handleBlur}
+              onClick={this.handleClick}
+              onFocus={this.handleFocus}
```
dgonzalezr, line 116: Shouldn't the `bq events` be used here? We have bqBlur, bqClick, and bqFocus in the `bq-button` component:
https://github.com/Endava/BEEQ/blob/main/packages/beeq/src/components/button/bq-button.tsx#L111-L118

```suggestion
              onBqBlur={this.handleBlur}
              onBqClick={this.handleClick}
              onBqFocus={this.handleFocus}
```

packages/beeq/src/components/page-title/bq-page-title.tsx
```diff
@@ -0,0 +1,170 @@
...
+    this.bqClick.emit(this.el);
+  };
+
+  private handleBlur = () => {
+    this.bqBlur.emit(this.el);
+  };
+
+  private handleFocus = () => {
+    this.bqFocus.emit(this.el);
+  };
```
dgonzalezr, line 92: We should stop propagation here, and probably instead of emitting flat events, to be more meaningful:
* `bqBackClick`
* `bqBackBlur`
* `bqBackFocus

packages/beeq/src/components/page-title/bq-page-title.tsx
```diff
@@ -111,6 +111,7 @@ export class BqPageTitle {
         <div class={{ flex: true, '!hidden': !this.haveBackNavigation }} part="back">
           <slot name="back">
             <bq-button
+              class="hover:[&::part(button)]:bg-transparent"
```
dgonzalezr, line 114: Why? 🤔

Cata1989: Initially, I thought that we did not need hover effect on the arrow as I did not see it in the Figma design, but since it's a `button`, it make sense to have this hover, so I will delete it. 😎

packages/beeq/src/components/page-title/_storybook/bq-page-title.stories.tsx
```diff
@@ -0,0 +1,112 @@
...
+  },
+};
+
+export const TitleBackSubtitle: Story = {
+  name: 'Title + Back + Subtitle',
+  render: Template,
+  args: {
+    'have-back-navigation': true,
+    title: 'Title',
+    subTitle: 'Sub-title',
```
dgonzalezr, line 75:
```suggestion
    'sub-title': 'Sub-title',
```

packages/beeq/src/components/page-title/_storybook/bq-page-title.stories.tsx
```diff
@@ -0,0 +1,112 @@
...
+  },
+};
+
+export const TitleBackActionsCustom: Story = {
+  name: 'Title + Back + Subtitle + Actions + Custom Divider',
+  render: Template,
+  argTypes: {
+    customDivider: { control: 'boolean', table: { disable: true } },
+    title: { control: 'text', table: { disable: true } },
+    subTitle: { control: 'text', table: { disable: true } },
```
dgonzalezr, line 99:
```suggestion
   'sub-title': { control: 'text', table: { disable: true } },
```

packages/beeq/src/components/page-title/_storybook/bq-page-title.stories.tsx
```diff
@@ -0,0 +1,112 @@
...
+    actions: { control: 'text', table: { disable: true } },
+  },
+  args: {
+    customDivider: true,
+    'have-back-navigation': true,
+    title: 'Title',
+    subTitle: 'Sub-title',
```
dgonzalezr, line 106:
```suggestion
    'sub-title': 'Sub-title',
```

packages/beeq/src/components/page-title/_storybook/bq-page-title.stories.tsx
```diff
@@ -0,0 +1,110 @@
...
+    docs: {
+      page: mdx,
+    },
+  },
+  argTypes: {
+    'have-back-navigation': { control: 'boolean' },
+    customDivider: { control: 'boolean', table: { disable: true } },
+    title: { control: 'text', table: { disable: true } },
+    'sub-title': { control: 'text', table: { disable: true } },
+    actions: { control: 'text', table: { disable: true } },
```
dgonzalezr, line 19: Please, try always to make a clear differentiation between the component's props and those used only for testing/documentation purposes.

```suggestion
    // Not part of the component's API, so we don't want to expose it in the controls panel
    customDivider: { control: 'boolean', table: { disable: true } },
    title: { control: 'text', table: { disable: true } },
    'sub-title': { control: 'text', table: { disable: true } },
    actions: { control: 'text', table: { disable: true } },
```

packages/beeq/src/components/page-title/bq-page-title.tsx
```diff
@@ -0,0 +1,183 @@
...
+              part="sub-title"
+            >
+              <slot name="sub-title" onSlotchange={this.handleSlotChange} />
+            </div>
+          </div>
+        </div>
+        {/* Divider */}
+        <div part="divider">
+          <slot name="divider">
+            <bq-divider class="mb-m block" stroke-color="ui--secondary" stroke-thickness="1"></bq-divider>
```
dgonzalezr, line 177:
```suggestion
            <bq-divider class="mb-m block" stroke-color="ui--secondary" stroke-thickness="1" />
```

packages/beeq/src/components/page-title/scss/bq-page-title.variables.scss
```diff
@@ -14,8 +12,6 @@
    * @prop --bq-page-title--font-weight-title - Page title font weight for title
    * @prop --bq-page-title--font-weight-subtitle - Page title font weight for subtitle
    */
-  --bq-page-title--paddingX: theme(spacing.xxl4);
-  --bq-page-title--paddingY: theme(spacing.xl);
```
dgonzalezr: The `paddingY` shouldn't be removed, only start/end padding. I will recommend using logical properties instead, meaning that you could set in the JSX, something like:

CleanShot 2024-06-06 at 11 04 22@2x

packages/beeq/src/components/page-title/scss/bq-page-title.variables.scss
dgonzalezr, line 1: On the other hand, there as a discussion with @endv-bogdanb where we found that using prefixes at the component level for CSS custom properties does not make much sense, because they are inside the component scope. That said, we could set our CSS props like this:

```css
  /* stylelint-disable custom-property-pattern */
  --paddingY: theme(spacing.xl);
  --subtitle-fontWeight: theme(fontWeight.medium);
  --subtitle-textColor: theme(textColor.text.secondary);
  --subtitle-textSize: theme(fontSize.l);
  --title-lineHeight: theme(lineHeight.regular);
  --title-fontWeight: theme(fontWeight.bold);
  --title-textColor: theme(textColor.text.primary);
  --title-textSize: theme(fontSize.xxl);
```

packages/beeq/src/components/page-title/scss/bq-page-title.variables.scss
dgonzalezr: More on logical properties here:
https://www.sitepoint.com/css-logical-properties-guide/

packages/beeq/src/components/page-title/scss/bq-page-title.variables.scss
Cata1989: Yes, make sense.

packages/beeq/src/components/page-title/bq-page-title.tsx
```diff
@@ -0,0 +1,174 @@
...
+          </div>
+          <div class="flex flex-grow flex-col gap-xs" part="content">
+            <div class="flex items-center gap-xs" part="title-suffix">
+              {/* Title */}
+              <div
+                class="title-font text-[length:--title-textSize] font-[--title-fontWeight] leading-[--title-lineHeight] text-[color:--title-textColor]"
+                part="title"
+              >
+                <slot />
+              </div>
```
dgonzalezr, line 141: To improve accessibility we should use `h1` for the page heading here:

```diff
-            <div
+            <h1
                class="title-font text-[length:--title-textSize] font-[--title-fontWeight] leading-[--title-lineHeight] text-[color:--title-textColor]"
                part="title"
              >
                <slot />
+            </h1>
-            </div>
```

packages/beeq/src/components/page-title/bq-page-title.tsx
```diff
@@ -0,0 +1,174 @@
...
+                  color="text--primary"
+                  name="arrow-left"
+                  size="24"
+                  weight="bold"
+                  part="icon"
+                  exportparts="base,svg"
+                />
+              </bq-button>
+            </slot>
+          </div>
```
dgonzalezr, line 132: After taking a better look into accessibility and how screen readers react to this new header, I think we should completely let the Back Navigation button up to the consumer:

``` suggestion
          {/* Back navigation button */}
          <div class={{ flex: true, '!hidden': !this.haveBackNavigation }} part="back">
            <slot name="back" />
          </div>
```

Then in Storybook, we can change a bit the Template:

```diff
  return html`
    <bq-page-title
      ?have-back-navigation=${args['have-back-navigation']}
      @bqBackClick=${args.bqBackClick}
      @bqBackFocus=${args.bqBackFocus}
      @bqBackBlur=${args.bqBackBlur}
    >
+      ${args['have-back-navigation']
+        ? html`
+           <bq-button appearance="link" slot="back">
+             <bq-icon
+                color="text--primary"
+                name="arrow-left"
+                weight="bold"
+                role="img"
+                title="Navigate back to the previous page"
+              ></bq-icon>
+            </bq-button>
+          `
+        : nothing}
      ${args.title} ${args['sub-title'] ? html`<div slot="sub-title">${args['sub-title']}</div>` : nothing}
      ${args.actions ? html`<div class="${actionsSlotClass}" slot="suffix">${args.actions}</div>` : nothing}
      ${customDivider}
    </bq-page-title>
  `;
```

By removing the predefined back button, we no longer need the events and their handlers 🙂 


https://github.com/Endava/BEEQ/assets/328492/5672e257-b7e2-484c-a5ea-1f48f371894d

dgonzalezr: You could also think about removing the `haveNavigationBack` property and handle the slot container display based in the slot content, same as you did in the sub-title and divider slot containers 😉

PR #1106, opened 2024-06-12
refactor(Date Picker): use CSS logical properties instead of physical values

Cata1989:
## Description

The goal of this PR is to modernize the CSS by incorporating logical properties for improved adaptability and support for different writing modes.

## Related Issue

Fixes #ISSUE_NUMBER

## Checklist

### DIFF
--- a/packages/beeq/src/components/date-picker/bq-date-picker.tsx
+++ b/packages/beeq/src/components/date-picker/bq-date-picker.tsx
@@ -485,7 +485,7 @@ export class BqDatePicker {
         </label>
         {/* Select date picker dropdown */}
         <bq-dropdown
-          class="bq-date-picker__dropdown w-full [&::part(panel)]:w-auto [&::part(panel)]:p-m"
+          class="bq-date-picker__dropdown is-full p-b-m [&::part(panel)]:is-auto [&::part(panel)]:p-i-m"
           disabled={this.disabled}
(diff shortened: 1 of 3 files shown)


packages/beeq/src/components/date-picker/bq-date-picker.tsx
```diff
@@ -485,7 +485,7 @@ export class BqDatePicker {
         </label>
         {/* Select date picker dropdown */}
         <bq-dropdown
-          class="bq-date-picker__dropdown w-full [&::part(panel)]:w-auto [&::part(panel)]:p-m"
+          class="bq-date-picker__dropdown is-full p-b-m [&::part(panel)]:is-auto [&::part(panel)]:p-i-m"
```
dgonzalezr, line 488: It seems that the addition of `p-b-m` is a mistake, maybe the intention was to set:
```suggestion
          class="bq-date-picker__dropdown is-full [&::part(panel)]:is-auto [&::part(panel)]:p-b-m [&::part(panel)]:p-i-m"
```
but if the padding remains the same all over, then `p-m` should be enough, no need to split in padding-block and padding-inline ;-)

dgonzalezr: I pushed a fix for it!

dgonzalezr (approved):

PR #1155, opened 2024-07-18
fix(Drawer): open/close animation to be in sync with the content direction

dgonzalezr:
## Description

This PR aims to solve an existing issue with the open/close animation behavior of the Drawer when the content direction switches from RTL to LTR, the animations are not working as expected since it's using `translate` which makes the drawer animates from/to the opposite side.

https://github.com/user-attachments/assets/9093d4c6-2cac-4382-8f5b-2cc27bd1e751

The fix removed the use of translation and relies on the inset-inline-start/end property, making it possible that animations work as expected regardless of the `direction="rtl | ltr"` set.

https://github.com/user-attachments/assets/d8a4a5e8-6e1a-4c9c-8deb-e2b15e2a3843

## Related Issue

N/A

## Checklist


Cata1989 (approved):
LGTM

Cata1989 (approved):

endv-bogdanb (approved):

PR #1643, opened 2026-04-17
chore(tests): migrate and improve component e2e tests

Morosan:
## Description
This PR updates and stabilizes the component test suite after the recent migrations, aligning the E2E coverage with the current component APIs and browser behavior.

Updated/migrated component coverage:
- `bq-badge`
- `bq-breadcrumb`
- `bq-breadcrumb-item`
- `bq-button`
- `bq-card`
- `bq-checkbox`
- `bq-date-picker`
- `bq-dialog`
- `bq-divider`
- `bq-drawer`
- `bq-dropdown`

Main changes:
- improved E2E coverage for interactive and public component behavior
- updated tests to reflect the migrated component contracts
- fixed assertions that no longer matched real browser behavior after the migrations
- kept the suite focused by adding coverage only where it increased confidence in the migrated implementation

Included updates:
- `bq-dialog`
  - added coverage for `show()` / `hide()`
  - added coverage for closing with `Escape`
- `bq-dropdown`
  - added coverage for `disabled`
  - added coverage for `bqSelect` behavior
  - added coverage for `keepOpenOnSelect`
- `bq-date-picker`
  - added coverage for the public `clear()` method
  - verified `bqClear` emission and value reset
- `bq-checkbox`
  - added coverage for `vClick()`, `vFocus()`, and `vBlur()`
- `bq-breadcrumb`
  - added coverage for `aria-current` on the last item
- `bq-drawer`
  - added coverage for closing with `Escape`
- `bq-button`
  - fixed the disabled test to avoid clicking a truly disabled native button
  - fixed reset behavior test to use `defaultValue`, matching native form reset behavior

## Checklist

- [x] I have reviewed my code.


dgonzalezr (changes requested):
Hey @Morosan, thanks a lot for putting this together — really solid work migrating all these tests to the new Vitest runner. The structure is clean and consistent across components, and the new additions (dialog show/hide, dropdown disabled, checkbox vClick/vFocus/vBlur, breadcrumb aria-current, date-picker clear, drawer Escape) are great improvements 💪

I went through each component and noted a few areas where we could push the coverage further. Nothing wrong with what's here — just opportunities to make it even more robust.

---

### `bq-badge`
- The `backgroundColor` and `textColor` props set inline CSS variables (`--bq-badge--background-color`, `--bq-badge--text-color`). Would be nice to have a test that verifies these get applied.
- The `@Watch` on `size` resets to `'small'` on invalid values — we could add an invalid prop fallback test (similar to what we already do in `bq-button` and `bq-divider`).
- Only single-digit content is tested (`2`). Multi-character content adds a `p-i-xs2` padding class — worth verifying.
- Minor: the "render a number" test manually queries assigned elements — we have `getTextContent` from `shared/utils/slot.ts` that handles this nicely and keeps things consistent.

### `bq-breadcrumb`
- The `label` prop drives `aria-label` on the `<nav>` element (defaults to `'Breadcrumbs'`). No test checks this attribute or a custom label value.
- The default `/` separator is cloned to all non-last items, but only the custom separator test covers multiple items. A test for the default separator behavior would round things out.

### `bq-breadcrumb-item`
- The component declares `bqFocus`, `bqBlur`, and `bqClick` events, but none are tested at the item level (only indirectly through the parent). Having isolated event tests here would be good.
- The `target` and `rel` props affect the rendered `<a>` tag but aren't verified.

### `bq-card`
- The `border` prop (`'none' | 'xs2' | ... | 'full'`) controls border-radius but isn't covered.
- The `@Watch` on `type` resets to `'default'` on invalid values — an invalid prop fallback test would be useful.

### `bq-checkbox`
- **Disabled state** is a significant gap — no test verifies that a disabled checkbox can't be checked, focused, or changed. For a form control this is important.
- Form association (`formAssociated: true`) brings `formResetCallback`, `setFormValue`, and `required` validation — none of these are tested.
- The `backgroundOnHover` prop adds a `has-background` CSS class — easy to verify.
- Setting `indeterminate` forces `checked = false` — only the visual mark is tested, not the state interaction.
- The slot text test manually queries assigned elements — could use `getTextContent` for consistency.

### `bq-date-picker`
- `disabled` state is uncovered — the component guards `bqFocus`, `bqBlur`, `clear()`, and all input handling when disabled.
- `bqFocus`, `bqBlur`, and `bqChange` events are never tested directly (only `bqClear` is).
- `disableClear` prop controls clear button visibility — not tested.
- `validationStatus` prop applies a CSS class and `aria-invalid` — not tested.
- `type` invalid fallback is not tested.
- The min/max clamping and clear tests are great — just needs more around disabled state and events.

### `bq-dialog`
- `disableCloseEscKeydown` — when `true`, Escape should NOT close the dialog. Would be good to test this alongside the existing Escape test.
- `disableCloseClickOutside` — not covered.
- `hideCloseButton` — not covered.
- `size` prop applies CSS classes (`small | medium | large`) — not covered.
- `show()` and `hide()` both abort if `defaultPrevented` — interesting edge case to cover.
- The slot text tests manually query assigned elements — `getTextContent` from `shared/utils/slot.ts` would simplify these.

### `bq-divider`
- The `dashed` prop toggles SVG stroke-dasharray — not tested.
- `orientation` is validated as a prop fallback, but no test checks that the `bq-divider--vertical` CSS class is actually applied when set to `'vertical'`.
- Title slot content (rendered between the two SVG lines) — not tested.
- The slot text test could use `getTextContent` for consistency.

### `bq-drawer`
- ⚠️ **Heads up on the Escape test** — `closeOnEsc` defaults to `false`, which means the component's `handleKeyDown` should return early and Escape should NOT close the drawer. The test dispatches Escape on `<bq-drawer open>` without setting `closeOnEsc`, so it might be asserting against behavior that shouldn't happen. Worth double-checking!
- `position` prop (`start | end`) applies CSS classes — not tested.
- `enableBackdrop` prop conditionally renders a backdrop — not tested.
- `closeOnClickOutside` — not tested.
- `show()`/`hide()` both abort if `defaultPrevented` — not tested.
- The title slot text test could use `getTextContent`.

### `bq-dropdown`
- The `bqOpen` event (emitted when `open` changes) is never spied on in any test.
- The `keepOpenOnSelect` test combines both the "close on select" and "keep open on select" cases in one `it()` block with two separate renders. Splitting into two tests would make failures easier to diagnose.
- Click-outside-to-close behavior is not tested.
- `sameWidth` and `panelHeight` props are not tested.

---

### General notes
- **`getTextContent` from `shared/utils/slot.ts`** — several tests manually query `assignedElements()[0].textContent`. This utility already handles that with trimming built in, so it'd simplify things across bq-dialog, bq-divider, bq-checkbox, bq-drawer, and bq-breadcrumb-item.
- **Disabled state** is the biggest coverage gap — especially for `bq-checkbox` and `bq-date-picker` where it guards all interactions.
- **Invalid prop fallback tests** are missing for `bq-badge`, `bq-card`, `bq-date-picker`, and `bq-dialog` — easy wins since the pattern is already established in `bq-button` and `bq-divider`.

Again, really nice work on the migration and thanks for the help! Let me know if you have questions on any of these 🙌

dgonzalezr (approved):

PR #1694, opened 2026-06-10
fix(Button): keep ghost button height consistent and prevent text selection

### LINKED ISSUE #1693: Bug: [Button] ghost variant height differs from default button and allows text selection on hover
### BEEQ package

beeq/core

### BEEQ version

1.12.3

### Current Behavior

When using bq-button with appearance="primary" and variant="ghost", the button ends up a few pixels taller than its non-ghost counterpart because of the border being added on top of the existing sizing.

This creates a visible inconsistency when a default/filled button and a ghost button are placed side by side.

Also, the cursor can become a text cursor when hovering the actual button content, and the label text can be selected.

### Expected Behavior

Ghost buttons should keep the same final height as their non-ghost counterparts across all sizes, without hardcoding fixed heights. The sizing should account for the border so a medium button still renders at 48px, and the same principle should apply to the other sizes as well.

In addition, the button surface inside the shadow DOM should prevent text selection so hovering/clicking behaves like a button, not like selectable text.

### Steps to Reproduce

N/A

### Code Reproduction URL

_No response_

### Your Environment

```shell

```

### Additional Information

_No response_

Morosan:
## Description
Keeps `bq-button` ghost variants visually aligned with their non-ghost counterparts by preserving the same final height across sizes, without hardcoding button heights.

This PR updates `packages/beeq/src/components/button/scss/bq-button.variables.scss` and `packages/beeq/src/components/button/scss/bq-button.scss` so button sizing accounts for border width through padding calculations, while ghost variants only override the border color instead of redefining border size. This ensures that default and ghost buttons can sit side by side without visible height differences.

It also adds `select-none` to `.bq-button` inside the shadow DOM so the button label does not become selectable and the interaction feels like a proper button after removing the host-level cursor rule.

Additionally, `packages/beeq/src/components/button/__tests__/bq-button.e2e.tsx` has been updated to validate the adjusted padding values and to verify that ghost buttons match the height of their non-ghost equivalents.

## Related Issue
Fixes https://github.com/Endava/BEEQ/issues/1693

## Documentation
No documentation changes were required for this update.

## Checklist
- [x] I have reviewed my code.

## Additional Notes
This change is limited to button styling and interaction polish. It does not modify the component API, but it does ensure more consistent visual alignment and button behavior across variants.

### DIFF
--- a/packages/beeq/src/components/button/scss/bq-button.variables.scss
+++ b/packages/beeq/src/components/button/scss/bq-button.variables.scss
@@ -24,14 +24,14 @@
   --bq-button--border-width: theme(borderWidth.m);
 
   --bq-button--small-paddingX: theme(spacing.xs);
-  --bq-button--small-paddingY: calc(theme(spacing.xs2) - theme(borderWidth.m));
+  --bq-button--small-paddingY: calc(theme(spacing.xs2) - var(--bq-button--border-width));
   --bq-button--small-font-size: theme(fontSize.m);
 
   --bq-button--medium-paddingX: theme(spacing.m);
-  --bq-button--medium-paddingY: calc(theme(spacing.s) - theme(borderWidth.m));
+  --bq-button--medium-paddingY: calc(theme(spacing.s) - var(--bq-button--border-width));
   --bq-button--medium-font-size: theme(fontSize.m);
 
   --bq-button--large-paddingX: theme(spacing.l);
-  --bq-button--large-paddingY: calc(theme(spacing.m) - theme(borderWidth.m));
+  --bq-button--large-paddingY: calc(theme(spacing.m) - var(--bq-button--border-width));
   --bq-button--large-font-size: theme(fontSize.m);
 }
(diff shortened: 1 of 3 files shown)


packages/beeq/src/components/button/scss/bq-button.variables.scss
```diff
@@ -21,17 +21,17 @@
   --bq-button--border-color: theme(borderColor.transparent);
   --bq-button--border-radius: theme(borderRadius.m);
   --bq-button--border-style: solid;
-  --bq-button--border-width: theme(borderWidth.0);
+  --bq-button--border-width: theme(borderWidth.m);
 
   --bq-button--small-paddingX: theme(spacing.xs);
-  --bq-button--small-paddingY: theme(spacing.xs2);
+  --bq-button--small-paddingY: calc(theme(spacing.xs2) - theme(borderWidth.m));
   --bq-button--small-font-size: theme(fontSize.m);
```
dgonzalezr: If the `--bq-button--border-radius` changes that might affect how the button looks.
We probably should:

```suggestion
--bq-button--small-paddingY: calc(theme(spacing.xs2) - var(--bq-button--border-radius));
```

Same for the rest.

dgonzalezr (approved):

